PR #51, opened 2015-01-09
added optional --user argument to IRkernel::installspec()

ihrke:
On my system, the call to 

~~~R
IRkernel::installspec()
~~~

failed with `Permission denied` because `ipython kernelspec install` installs the kernel globally by default. I added an option `user` which allows to pass the `--user` flag to the call to `ipython kernelspec install`.


ihrke:
I'll open a new PR just with the change described here. This PR mixed two different issues.

PR #53, opened 2015-01-09
added optional --user argument to IRkernel::installspec()

ihrke:
On my system, the call to 

~~~R
IRkernel::installspec()
~~~

failed with `Permission denied` because `ipython kernelspec install` installs the kernel globally by default. I added an option `user` which allows to pass the `--user` flag to the call to `ipython kernelspec install`.

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -200,7 +200,7 @@ main <- function(connection_file="") {
 #'
 #'@param user Install into user directory ~/.ipython or globally?
 #'@export
-installspec <- function(user=F) {
+installspec <- function(user=T) {
     srcdir = system.file("kernelspec", package="IRkernel")
     user_flag=ifelse(user, "--user", "")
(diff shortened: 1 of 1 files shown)


R/kernel.r
```diff
@@ -198,9 +198,11 @@ main <- function(connection_file="") {
 
 #'Install the kernelspec to tell IPython (>= 3) about IRkernel
 #'
+#'@param user Install into user directory ~/.ipython or globally?
 #'@export
-installspec <- function() {
+installspec <- function(user=F) {
```
takluyver: Let's make the default True - since R packages typically install to a per-user directory, it makes sense for the kernelspec to follow suit. And that way, there's no need to update the README.

ihrke: Agreed. I used F because I did not want to change the default behavior.

takluyver: No problem. Actually, the default was originally to do a per-user install, until a change I made in IPython that switched the default. So this is actually bringing the default back to what I originally intended.

PR #63, opened 2015-03-07
set .GlobalEnv during execution

### LINKED ISSUE #61: save.image doesn’t work
it doesn’t save everything. the docs say it’s the same as

```r
save(list = ls(all.names = TRUE), file = ".RData", envir = .GlobalEnv)
```

something here goes wrong. is it `.GlobalEnv` or `ls()`?

flying-sheep:
~~maybe~~ definitely fixes #61


R/execution.r
```diff
@@ -37,9 +37,19 @@ namedlist <- function() {
 # This function copied from testthat, by Hadley Wickham. MIT licensed.
```
takluyver: For completeness, this should now say 'originally copied'

takluyver: Actually, I'll just merge and change that myself.

takluyver:
I see it's gone from 'maybe' to 'definitely' now - thanks for testing.

PR #64, opened 2015-03-08
Autocompletion

### LINKED ISSUE #5: Tab completion
Handle a [complete_request](http://ipython.org/ipython-doc/dev/development/messaging.html#completion) to make tab completion work.

flying-sheep:
depends on #63, fixes #5

maybe in the far future we can use [this monster](https://github.com/rstudio/rstudio/blob/master/src/cpp/session/modules/SessionRCompletions.R) instead…

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -99,9 +99,41 @@ handle_shell = function() {
         kernel_info(msg)
     } else if (msg$header$msg_type == "history_request") {
         history(msg)
+    } else if (msg$header$msg_type == "complete_request") {
+        complete(msg)
     } else {
         print(c("Got unhandled msg_type:", msg$header$msg_type))
     }
+},
+
+complete = function(request) {
+    # 5.0 protocol:
+    #code <- request$content$code
+    # once this is supported, we’ll need to change the
+    # .assignEnd call because it expects in-line pos
+    
+    # 4.0 protocol:
+    code <- request$content$line
+    
+    c.info <- with_top_env(executor$userenv, {
+        utils:::.assignLinebuffer(code)
+        utils:::.assignEnd(request$content$cursor_pos)
+        utils:::.guessTokenFromLine()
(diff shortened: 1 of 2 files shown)


takluyver:
Awesome, thanks! It looks like this should be easy to adapt to the msgspec5 request - I hope to make that change soon.

R/kernel.r
```diff
@@ -99,11 +99,43 @@ handle_shell = function() {
...
+    code <- request$content$line
+    
+    c.info <- with_top_env(executor$userenv, {
+        utils:::.assignLinebuffer(code)
+        utils:::.assignEnd(request$content$cursor_pos)
+        utils:::.guessTokenFromLine()
+        utils:::.completeToken()
+        
+        c(list(comps = utils:::.retrieveCompletions()),
+          utils:::.guessTokenFromLine(update = FALSE))
```
takluyver: I was a bit puzzled by what was going on here - is it worth a comment to explain that .guessTokenFromLine with update=FALSE returns the start and token fields? (it also surprised me that `update` changes the return type).

flying-sheep: ha! yeah sorry: that’s exactly the perfect case for a comment.

yes: all those functions act on an environment called `.CompletionEnv`, where they set a few variables.

maybe we could pull the variables out of the `.CompletionEnv` directly, but i found that calling this function with `update = FALSE` does exactly that. (and i’ve not found how to access `.CompletionEnv`)

lucacerone:
I tried it on my Jupyter install with python 2 and it seems to work like a champ!

flying-sheep:
@takluyver would you pull it?

i addressed your concern about comments

takluyver:
Thanks! Sorry it took me a while to get to this - this week has been the IPython dev meeting, so I've been pretty busy.

flying-sheep:
no worries :)

great this got in!

PR #71, opened 2015-03-19
Run user code in .GlobalEnv

takluyver:
I suspect this is causing many of the issues where executing code doesn't work exactly like the R interactive shell. I hadn't realised how specially R treats the global environment.

@flying-sheep , can you see any obvious problems with this approach? I've checked that the example notebooks still run OK, but I'm not an R expert.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -34,27 +34,8 @@ namedlist <- function() {
     return(setNames(list(), character(0)))
 }
 
-# This function originally copied from testthat, by Hadley Wickham. MIT licensed.
-# https://github.com/cran/testthat/blob/9e78f643d5b1c5d5882849508772ecbe980d3ac9/R/test-package.r
-with_top_env <- function(env, code) {
-    old.env <- .BaseNamespaceEnv$.GlobalEnv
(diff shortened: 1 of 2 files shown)


flying-sheep:
i’m not by any means an R expert.

what i gather is that `.GlobalEnv` is always the user environment. so i *think* it will work like this:

1. `R -e 'IRkernel::main()' --args $connectionfile` is called
2. this starts R with a new, empty `.GlobalEnv`
3. everything we assign to is in the `IRkernel` package namespace
4. everything runs normally with `.GlobalEnv` on top of the name search stack (like in a normal R session)

so this sounds pretty good. everything behaves like a normal R session AFAICS.

takluyver:
Thanks, that matches my understanding, so I'm going to merge this and see how it works with people using it.

`displayenv` should be the namespace used for the `IRdisplay` package - it's set at the top of execution.r:

```
displayenv = environment(display)
```

We use this to reassign `IRdisplay::base_display()` to a function that knows how to send a display message to the frontend.

flying-sheep:
OK, thanks!

PR #72, opened 2015-03-21
Switch to msgspec 5

takluyver:
A few things were renamed to be less Python specific. The biggest changes were to completion; now the entire cell is sent for completion. This code finds the correct line within that and passes it to R's completion machinery.

@flying-sheep , could you have a look at the completion machinery to see if I've done anything obviously wrong with it? Thanks.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -41,7 +41,7 @@ Executor = setRefClass("Executor",
 execute = function(request) {
   send_response = kernel$send_response
   send_response("status", request, 'iopub', list(execution_state="busy"))
-  send_response("pyin", request, 'iopub',
+  send_response("execute_input", request, 'iopub',
                 list(code=request$code, execution_count=execution_count))
 
   silent = request$content$silent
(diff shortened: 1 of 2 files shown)


flying-sheep:
looks good!

i have a rather “cute” alternative to the line exctraction code, which is a little faster, but that shouldn’t make a big difference except for huge cells

```r
getline = function(multiline, pos) {
    # no \n before start and after end, so we “add them”
    all.pos = c(0, gregexpr('\n', multiline)[[1]], nchar(multiline) + 1)
    # closest \n to pos, so either the one ending its line or the line before
    i = which.min(abs(all.pos - pos))

    # “≥” means a \n is interpreted as part of the line it ends
    leftshift = as.integer(all.pos[[i]] >= pos)
    start  = all.pos[[i - leftshift    ]] + 1  # position after left \n
    end    = all.pos[[i - leftshift + 1]] - 1  # position before right \n

    list(line   = substr(multiline, start, end),
         start  = start,
         end    = end,
         length = end - start + 1)
}
```

needs unit tests if you decide to use it

takluyver:
Yeah, it makes sense that that would be more efficient, if perhaps a bit harder to understand. I'm going to merge this as is - you can open a PR if you think it's worth changing the line extraction.

We're hoping to have a little tool to easily test kernels at the messaging layer (see ipython/ipython#7646) that should be useful for checking IRkernel.

Issue #75, opened 2015-04-06
 IRkernel::installspec() is broken on windows


mikecroucher:
> IRkernel::installspec()
Traceback (most recent call last):
  File "C:\Anaconda\Scripts\ipython-script.py", line 5, in <module>
    sys.exit(start_ipython())
  File "C:\Anaconda\lib\site-packages\IPython\__init__.py", line 120, in start_ipython
    return launch_new_instance(argv=argv, **kwargs)
  File "C:\Anaconda\lib\site-packages\IPython\config\application.py", line 574, in launch_instance
    app.start()
  File "C:\Anaconda\lib\site-packages\IPython\terminal\ipapp.py", line 367, in start
    return self.subapp.start()
  File "C:\Anaconda\lib\site-packages\IPython\kernel\kernelspecapp.py", line 142, in start
    return self.subapp.start()
  File "C:\Anaconda\lib\site-packages\IPython\kernel\kernelspecapp.py", line 83, in start
    replace=self.replace,
  File "C:\Anaconda\lib\site-packages\IPython\kernel\kernelspec.py", line 195, in install_kernel_spec
    shutil.copytree(source_dir, destination)
  File "C:\Anaconda\lib\shutil.py", line 171, in copytree
    names = os.listdir(src)
WindowsError: [Error 3] The system cannot find the path specified: u'C:/Program\\*.*'

If you suspect this is an IPython bug, please report it at:
    https://github.com/ipython/ipython/issues
or send an email to the mailing list at ipython-dev@scipy.org

You can print a more detailed traceback right now with "%tb", or use "%debug"
to interactively debug it.

Extra-detailed tracebacks for bug-reporting purposes can be enabled via:
    c.Application.verbose_crash=True

Warning message:
running command 'ipython kernelspec install --replace --name ir --user C:/Program Files/R/R-3.1.3/library/IRkernel/kernelspec' had status 1

takluyver:
Probably need some quoting because of the space in the path. PRs welcome, or I'll try to find time to deal with it.

mikecroucher:
Yep - that was exactly the problem. PR submitted and I see you merged it - thank you!

PR #76, opened 2015-04-06
Fix for paths containing spaces on Windows

mikecroucher:
This fixes issue #75 
I haven't been able to check that I haven't broken anything on Mac or Linux yet but I suspect that the quotes won't affect anything there.


takluyver:
Oh,  thanks, I read #75 before this. Looks fine to me, I'll ping you if anyone reports that it's broken ;-)

PR #81, opened 2015-04-12
Stressing importance of active IPython during install

drorata:
It is important that the instance of `ipython` that is assigned to the shell will be the right one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,10 @@ you'll lose all your variables if it crashes.
       IRkernel::installspec()
       ```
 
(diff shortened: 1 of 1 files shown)


takluyver:
I don't think it should be a problem if ipython is in a virtualenv - the issue you hit was specifically because it was finding an *older* version of IPython. Hopefully, that sort of confusion should decrease quickly as IPython >=3 replaces older versions. And the relevant line already mentions "Only if you have IPython 3 or above installed".

Thanks!

Issue #59, opened 2015-03-05
add setting for using HTML print function


flying-sheep:
like shown [here](http://nbviewer.ipython.org/gist/deeplook/5162445), ipython/jupyter has display hooks that can be registered with display formatters (e.g. png, html)

it would be nice if IRDisplay and IRKernel would work together so that

1. we define a S3 or S4 generic (e.g. `jupyterdisplay.default`)
2. IRDisplay defines it for some classes (e.g. `jupyterdisplay.matrix`)
3. users can override it

takluyver:
I think that makes sense - the default would just send the text/plain representation. We'd call `jupyterdisplay(obj)` on expressions at the end of a cell (the 'value' handler in evaluate), and packages could override it to display richer data for their objects, right? I'm not especially familiar with R's class mechanisms.

flying-sheep:
well, my idea was that

1. directly calling it will work like you described
2. `options(jupyter.display = 'rich')` would make it the default function to be called instead of `print`

i gather that `print` simply outputs to stdout (and returns its argument invisibly), and if something gets returned but not assigned in a REPL, then `print` is called on it.

---

also we basically already do something like this with plots, and have to think about what else should be treated specially (i.e. where don’t we just display the stdout of `print` even if `getOption('jupyter.display') == 'text'`?)

---

finally we should probably replace `set_plot_options` with `options(jupyter.plot = list(...))` if we’re going to add more options.

flying-sheep:
we should also add a class that allows HTML passthrough (like the `IPython.display.HTML` class).

then we could e.g. do:

```r
data.frame(
	escaped=c('<', '&', '<script>doEvil()</script>'),
	unescaped=c(as.html('<a href="..."></a>'), html('&amp;', '2<sup>nd</sup> cell')),
	factor=as.factor(........),
	stringsAsFactor=FALSE)
```

and only columns with `class(column) == 'html'` wouldn’t be escaped.

flying-sheep:
also we need to treat output differently from displaying.

normally things that are non-invisibly returned (e.g. by calling a function or typing a variable name without assigning it to a name) are `print`ed in R.

this already works: `print(foo)` does the same as `foo`, except that the latter produces an  `Out [1]:` thingy next to it (`execute_result`) and the former doesn’t (`display_data`)

---

i think we should do it like this:

1. define some rich display option
2. if it’s not set, print and output behaves normally (monospace string representation without and with `Out[1]`, respectively)
[...]

PR #82, opened 2015-04-13
Display changes

### LINKED ISSUE #73: Remove set_plot_options
…or deprecate and hide it from the docs.

R has a notion of the “default graphics device”, i.e. `getOption('device')`, which is either a string containing a function name or a function.

we could leverage this in two ways:

1. use it with custom options:

	```r
	dev <- getOption('device')
	if (!is.function(dev)) {
		dev <- get(dev)
	}
	do.call(dev, as.list(options('jupyter.plot')))
	```

2. require the user to override the defaults by setting it to a function:

	```r
	#our initialization
	options(device = function(file.name, width = 7, height = 7,...) {
		png(file.name, width = width, height = height, ...)
	})
	
	…
	
	dev <- getOption('device')
	if (!is.function(dev)) {
		dev <- get(dev)
	}
	dev()
	```

---

the only problem: we want to allow different mime types (PNG, JPG, SVG) for plots: if it’s a function, we can’t derive the type from the option.

flying-sheep:
i think this should be the way everyone’s happy with.

depends on ~~IRkernel/IRdisplay#3’s function `display_alternatives` and~~ [IRkernel/repr](https://github.com/IRkernel/repr) (in order to display multiple alternative formats of one output, and to create those formats, respectively)

sorry for the mess of commits, i brought the notebooks up to date too. relevant changes should start [here](https://github.com/IRkernel/IRkernel/pull/82/files#diff-6eae7f9cccd978b5b4d819baa7a34b1e)

this fixes #73 and lays foundations for #59 (by depending on repr)


R/execution.r
```diff
@@ -83,12 +62,13 @@ execute = function(request) {
...
-      replayPlot(plotobj)
-      dev.off()
-      display_png(file=tf)
+  send_plot <- function(plotobj) {
+      params <- list()
+      for (mime in getOption('jupyter.plot_mimetypes')) {
+          r <- mime2repr[[mime]](plotobj)
+          params[[mime]] <- if (is.raw(r)) base64encode(r) else r
+      }
+      do.call(display_alternatives, params)
```
JanSchulz: Ok, I was going to argue that this should send a `execution_reply` but in python this also uses the `display` function :-/

But this then means that the kernel depends on the display package (due to the useage in sending plots) and the display package depends on the kernel (due to the actual sending of the `display_data` reply). I'm not sure if that's a problem in R.

takluyver: This is already the case - IRdisplay provides the display API, that I also want to make usable from the R magic in a Python notebook, so the same code works in both. However, it doesn't know how to actually display the data - it needs to be run by something that gives it a base display function to send the data. IRkernel does this. It's not very nice, but it seems to work.

I'm still not convinced that display_alternatives is necessary, though.

flying-sheep:
so regarding #83:

you call that a hack? [1009088](https://github.com/IRkernel/IRkernel/commit/10090886efb30cc4948772b7e3b61ba9cfae4d7e) is a hack!

A hack

flying-sheep:
btw. i think that this is now independent from IRkernel/IRdisplay#3, right?

so we can pull this in and bikeshed over the IRdisplay apis the following weeks ;)

takluyver:
Yes, I think so - I tested it without that PR (using #83).

I also want to think a bit about repr APIs, though - I'm a bit confused as to why repr has a class, rather than just a set of methods returning string/raw vectors.

flying-sheep:
> I also want to think a bit about repr APIs, though - I'm a bit confused as to why repr has a class, rather than just a set of methods returning string/raw vectors.

let’s discuss this in IRkernel/repr#1

PR #83, opened 2015-04-13
Don't use display_alternatives

takluyver:
@flying-sheep this is a mini-PR against #82 to illustrate why I think `display_alternatives` is unnecessary.

`unclass` is admittedly a hack, but I don't have time to work out what's going on with the 'repr' class. I suspect that could be simplified too.


flying-sheep:
you’re right, but i’ll change this locally since more changes are already in the pipeline, eagerly awaiting the review they’re based upon ;)

PR #84, opened 2015-04-14
Added repr-based responses and corresponding options

### LINKED ISSUE #59: add setting for using HTML print function

flying-sheep:
fixes #59

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -92,12 +92,20 @@ execute = function(request) {
     handle_message = identity
     handle_warning = identity
   } else {
-    handle_value = function (obj) {
-        data = list()
-        data['text/plain'] = paste(capture.output(print(obj)), collapse="\n")
-        send_response("execute_result", request, 'iopub',
-                  list(data=data, metadata=namedlist(),
-                  execution_count=execution_count))
+    handle_value <- function (obj) {
+        data <- list()
+        if (getOption('jupyter.rich_display')) {
+            for (mime in getOption('jupyter.result_mimetypes')) {
+                r <- mime2repr[[mime]](obj)
+                if (!is.null(r)) data[[mime]] <- r
+            }
+        } else {
+            data[['text/plain']] <- repr_text(obj)
+        } 
+        send_response('execute_result', request, 'iopub', list(
+            data = data,
+            metadata = namedlist(),
+            execution_count = execution_count))
     }
     stream = function(output, streamname) {
         send_response("stream", request, 'iopub',
--- a/R/execution.r
+++ b/R/execution.r
@@ -22,11 +22,6 @@ plot_builds_upon <- function(prev, current) {
     return((lcurrent >= lprev) && (identical(current[[1]][1:lprev], prev[[1]][1:lprev])))
 }
 
-# needed to easily encode reprs as json.
-setOldClass('repr')
-asJSON <- jsonlite:::asJSON
-setMethod('asJSON', 'repr', function(x, ...) jsonlite:::asJSON(structure(x, class = NULL, repr.format = NULL), ...))
-
 Executor = setRefClass("Executor",
             fields=c("execution_count", "payload", "err", "interrupted", "kernel",
                      "last_recorded_plot"),
--- a/R/options.R
+++ b/R/options.R
@@ -33,6 +33,7 @@ opt.defaults <- list(
         'image/svg+xml'),
     jupyter.plot_mimetypes = c(
         'image/png',
+        'image/jpeg',
         'application/pdf',
         'image/svg+xml'))
 
--- a/R/options.R
+++ b/R/options.R
@@ -16,13 +16,22 @@ get_plot_options <- function() {
     return(as.list(plot.opts))
 }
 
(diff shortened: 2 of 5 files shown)


R/options.R
```diff
@@ -16,13 +16,22 @@ get_plot_options <- function() {
...
-    'image/png',
-    'application/pdf',
-    'image/svg+xml')
+opt.defaults = list(
+    jupyter.rich_display = TRUE,
+    jupyter.result_mimetypes = c(
+        'text/plain',
+        'text/html',
+        'text/markdown',
+        'text/latex'),
```
takluyver: Why wouldn't we try all of them? It could make sense to have e.g. a png representation of something other than a plot.

R/execution.r
```diff
@@ -92,12 +92,20 @@ execute = function(request) {
...
-                  execution_count=execution_count))
+    handle_value <- function (obj) {
+        data <- list()
+        if (getOption('jupyter.rich_display')) {
+            for (mime in getOption('jupyter.result_mimetypes')) {
+                r <- mime2repr[[mime]](obj)
+                if (!is.null(r)) data[[mime]] <- r
+            }
+        } else {
+            data[['text/plain']] <- repr_text(obj)
```
takluyver: For my own curiosity, what's the difference between setting `data['text/plain']` and `data[['text/plain']]`? In a quick test, the result appears to be the same, but I seem to remember they're not equivalent.

R/options.R
JanSchulz: ipython also has a knob to dis/enable some of the formats. It's used by the mpl inline backend to set the send back image formats.

JanSchulz: Anyway: the way the system there works, it really tries all of the enabled ones, so even for plots, the `repr_markdown`would be tried (and return nothing)

takluyver: Sorry, I should have been clearer: I agree with having an option to configure this, but I'd put all the formats we know IPython can display in there by default - I think that just means adding image/png and image/svg+xml to this list.

flying-sheep: [and many more](http://ipython.org/ipython-doc/stable/api/generated/IPython.core.formatters.html#IPython.core.formatters.format_display_data). i’ll add them.

JanSchulz:
Whatever comes out of these discussion: I think it would be good to add a blogpost like document to the default Jupyter documentations where this system is explained for "kernel developers", so that other can use that to model their own system after it...

R/options.R
flying-sheep: do you think i should do the same for `jupyter.plot_mimetypes`?

how does ipython handle this?

R/execution.r
flying-sheep: it’s more clear.

[[]] means: get/set one element

[] means: get/set a slice

this is important when e.g. slicing something with names:

```r
> l <- list(a = 1, b = 2)
> l[1]
$a
[1] 1
> l[[1]]
[1] 1
```

the first one will return a list that is a slice of `l`, the second one will return an element.

also consider this:

```r
> # v is a vector
> # i should be a single integer but due to a bug is c(1,2)
> x <- v[[i]]
Fehler in l[[i]] : Indizierung außerhalb der Grenzen
> x <- v[i]
> # x is expected to be one value here, but is a vector of length 2 instead
```

the first one shows the error, the second one silently propagates it.

flying-sheep: in our example, someone could do`options(jupyter.result_mimetypes = list(c('text/plain', 'text/html'), 'text/latex'))`.

`mime2repr[mime]` would then be `list(repr_text, repr_html)` (and calling it would throw an error, true, but better be clear everywhere)

R/options.R
JanSchulz: Ipython makes no difference between plotting and displaying anything else, the "knob" to influence the plotting formats actually just is a convenience function, which (de-)activates the image formatter.

flying-sheep:
since the discussion is hidden now:

the default for results is now [everything](http://ipython.org/ipython-doc/stable/api/generated/IPython.core.formatters.html#IPython.core.formatters.format_display_data).

do you think i should do the same for `jupyter.plot_mimetypes`?

JanSchulz:
I think the problem here is that if the plotting option is just "setting an option like "jupyter.plot_mimetypes" to a list", we miss out any validation/"translation" to the normal display formatter like the ipython equivalent does. Or is there a way to validate options when it is set via `options(...)`?

flying-sheep:
we could use the [settings](http://cran.r-project.org/web/packages/settings/vignettes/settings.html) package.

flying-sheep:
so, any objections to pulling this in?

R/execution.r
takluyver: Thanks, that makes sense.

takluyver:
I think this looks good.

PR #86, opened 2015-04-16
got rid of display_alternatives import

khondula:
Hi! I was running into this error installing IRkernal: 
Error : object ‘display_alternatives’ is not exported by 'namespace:IRdisplay'

Not sure if this will cause any other problems, but I was able to complete the installation by deleting the importFrom(IRdisplay,display_alternatives)


flying-sheep:
forgot this. damn.

flying-sheep:
now it also won’t reappear by recreating the namespace using roxygen :yum:

Issue #54, opened 2015-01-19
Windows Support


JanSchulz:
It would be nice to prepare a package for rzmq and irkernel, so that windows user can install them without setting up a mingw environment. Or upload them to cran :-)

I did get it to install the 64 bit version:

* `install.packages('RCurl')` and `library(devtools)`
* compile zeromg with a mingw by following this howto: http://zeromq.org/docs:windows-installations#toc2. I changed `Makefile` to remove perf and test from SUBDIRS to make compiling faster. This builds a 64bit libzmq.dll, but no 32 bit... I couldn't get it to compile using the rtools (setting the path to c:/rtools/bin and using `C:\data\downloads\zeromq-4.0.5\zeromq-4.0.5>sh configure --prefix=C:/rtools/zeromq build_alias=x86-mingw32 "CFLAGS=-m32" "CXXFLAGS=-m32" "LDFLAGS=-m32"`) and also not the 32bit version with the above mingw (passing in  ` "CFLAGS=-m32" "CXXFLAGS=-m32" "LDFLAGS=-m32"` to configure...) 
* clone armstrtw/rzmq and git submodule init/update to get `inst\cppzmq\zmq.hpp`. Also copied `zmq.h` from the compiled zeromq folder into that dir, because I didn't know how to find the usr/local/include folder in rtools...
* copied libzmq.dll together with libgcc_s_sjlj-1.dll and libstdc++-6.dll into a lot of subfolders of C:\Rtools\gcc-4.6.3 (*\bin, *\lib; not sure which was the correct one in the end) until the following worked:
* in rzmq folder: `C:/PROGRA~1/R/R-31~1.2/bin/x64/R --vanilla CMD INSTALL  %CD% --library="C:/portabel/Rlibs" --no-multiarch`
* `install_github("takluyver/IRdisplay")`
* clone takluyver/IRkernel and in irkernel again:  `C:/PROGRA~1/R/R-31~1.2/bin/x64/R --vanilla CMD INSTALL  %CD% --library="C:/portabel/Rlibs" --no-multiarch`
* run `C:/PROGRA~1/R/R-31~1.2/bin/x64/R` and then `IRkernel::installspec()`
[...]

Issue #90, opened 2015-04-18
Please add a repo with all packages (including binaries) as a gh-page


JanSchulz:
I've added such a repository at http://janschulz.github.io/IRkernel/ but only for windows (R 3.1 and 3.2) and source packages. This is pullabale via the 'gh-pages' branch in https://github.com/JanSchulz/IRkernel.git (I can't create a PR against a nonexisting branch in this Repo :-( ).

For a user, this is useable as `install.packages(c("rzmq","repr","IRkernel","IRdisplay"), repos="http://janschulz.github.io/IRkernel")`

My script to generate the windows version looks like this (it has two dirs for the IRkernel: IRkernel with master checked out and IRkernel-ghpage with the 'gh-pages' branch checked out; rzmq is checked out with my PR in https://github.com/armstrtw/rzmq/pull/16; I use drat as in https://github.com/eddelbuettel/drat/pull/16, which has binary support for windows and macosx, but the latter untested):

```
set OLDPATH=%path%
set path=c:\Rtools\bin;c:\Rtools\gcc-4.6.3\bin;C:\Program Files\R\R-3.1.2\bin\x64;%OLDPATH% 
echo "building for R3.1"

rem @echo off
::setlocal EnableDelayedExpansion
set old=%CD%
cd %~dp0

R -e "R.Version()"
pause

for %%G IN (rzmq,repr,IRKernel,IRDisplay) do (
   echo %%G
   R CMD build %%G
   rem R CMD check %%g_*.tar.gz
   R CMD INSTALL --build --compile-both %%G
   )

pause

::goto :EOF

for /f "tokens=*" %%G in ('dir /b "*.zip"') do (
   echo %%G
   R --slave -e "drat:::insert('%%~nG%%~xG', repodir='IRkernel-ghpages')"
   move /y %%G packages\3.1
   )

for /f "tokens=*" %%G in ('dir /b "*.tar.gz"') do (
   echo %%G
   R --slave -e "drat:::insert('%%~nG%%~xG', repodir='IRkernel-ghpages')"
   move /y %%G packages\3.1
   )

pause

set path=c:\Rtools\bin;c:\Rtools\gcc-4.6.3\bin;C:\Program Files\R\R-3.2\bin\x64;%OLDPATH% 
echo "building for R3.2"

R -e "R.Version()"
pause

rem @echo off
::setlocal EnableDelayedExpansion
set old=%CD%
cd %~dp0

[...]

PR #95, opened 2015-04-23
Update Readme with new information how to install

JanSchulz:
Make installing from the supplied packages the new default and
mention that only IPython 3.1 and above is supported.

Closes: #54 # windows support

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,53 @@ This is still experimental. Your code should be safe,
 since IPython handles saving and loading notebooks in another process, but
 you'll lose all your variables if it crashes.
 
-## Installing
+## Requirements
+
+To run IPython with an R kernel, you need at least the following:
+
+* [IPython](http://ipython.org/), version 3.0 or later. If you already have a 
+  Python environment set up, install IPython using your preferred tools. If 
+  not, installing [Anaconda](http://continuum.io/downloads) is the quickest 
+  way to get everything you need. Earlier versions of IPython are now 
+  unsupported.
+* A current [R installation](http://www.r-project.org/).
+
+Installing from source needs additionally header files (see below).
+
+## Installing via supplied binary packages (Windows)
+
+We provide windows binary packages of all the needed packages. The packages 
+and the kernel spec can be installed with the following lines in an R console: 
+
+```r
+install.packages(c("rzmq","repr","IRkernel","IRdisplay"), repos="http://irkernel.github.io/")
+IRkernel::installspec()
+```
+
+To update packages, you have to either add the repo to your default ones in your R startup 
+file and update afterwards as normal:
+
+```r
+r <- getOption("repos")
+r["IRkernel"] <- "http://irkernel.github.io/"
+options(repos = r)
+```
+
+Or use the `repo` option to `update.packages()` directly:
+
+```r
+update.packages(repos="http://irkernel.github.io/")
+```
+
+Please note that during the initial development, these packages can be updated
+without changing the version number, which prevents updating via `update.packages()`. 
+In that case, updated packages can be installed by re-running the above 
+`install.packages()` line.
+
+## Installing from source (Linux/Mac)
 
 1. You'll need zmq development headers to compile rzmq, as well curl headers
    for R devtools.
-
   * **Ubuntu/Debian**
 
     ```bash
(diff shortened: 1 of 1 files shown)


takluyver:
IPython 3.0 does work, as far as I know - the problem you encountered with knitpy only affects other applications starting kernels through the IPython API, not IPython applications. (And it's probably possible to work around, but since 3.1 is already out, I don't think it's worth it).

README.md
```diff
@@ -4,7 +4,48 @@ This is still experimental. Your code should be safe,
...
+  
+We provide source and windows binary packages of the needed package. They can be installed 
+with the following lines in an R console: 
+
+```r
+install.packages(c("rzmq","repr","IRkernel","IRdisplay"), repos="http://irkernel.github.io/")
+IRkernel::installspec()
+```
+
+On Mac, you probably need to `install.packages(... , type="source")`.
```
takluyver: Probably? Do we know what happens if you don't?

README.md
```diff
@@ -4,7 +4,48 @@ This is still experimental. Your code should be safe,
...
+
+If this does not work, please see below for installing from source.
+  
+Please note that during the initial development, these packages can be updated
+without changing the version number, so updated packages can be installed by
```
takluyver: I hope we don't have to do this, though perhaps it's a sensible disclaimer for now.

README.md
JanSchulz: I know I have to use it on windows if the repo only has source packages (windows will try binaries first/only and raise something about no packages found if there are no binary packages or the PACKAGE file for binaries is missing). I'm not sure what mac does.

README.md
JanSchulz: To update from that repo we would have to add the repo to `getOption("repos")`:

```
r <- getOption("repos")
r["IRkernel"] <- "http://irkernel.github.io/"
options(repos = r)
```

README.md
JanSchulz: Anyway: I think the most sensible thing would be to have someone compile the mac packages and install them via drat in that repo...

README.md
```diff
@@ -4,7 +4,48 @@ This is still experimental. Your code should be safe,
 since IPython handles saving and loading notebooks in another process, but
 you'll lose all your variables if it crashes.
 
-## Installing
+## Installing 
+   
+Requirements:
+
+* [IPython](http://ipython.org/), version 3.1 or later. If you already have a
```
JanSchulz: Ok, will make that "3.0 or later"

takluyver:
On any platforms where we're installing source packages, we will still need to have the ZMQ headers installed, e.g. `libzmq3-dev` on Debian. Otherwise it won't be possible to compile rzmq.

JanSchulz:
Ok, so do we know anybody with with R and a Mac? If so we could include mac packages and the Readme would become "if you are on win/mac, do this, and on linux install the zmq headers and do that." :-)

flying-sheep:
@noamross from #96 apparently has one :)

sckott:
i have a mac too - should I test something?

noamross:
`install.packages(c("rzmq","repr","IRkernel","IRdisplay"), repos="http://irkernel.github.io/", type="source")` worked for me

JanSchulz:
@noamross Can you do some package building? Basically checkout all the above packages and then do:

```
# for g in "rzmq","repr","IRkernel","IRdisplay":
R CMD build %G%
R CMD check %G%_*.tar.gz
R CMD INSTALL --build --compile-both %G%
```
(Sorry, my bash is rusty, no idea about the loop :-( )

This should produce a lot of *tar.gz (the source packages) and *.tgz (the mac binary packages). The latter could be added to the repo via drat (see here for binary support: https://github.com/eddelbuettel/drat/pull/16)

noamross:
IRkernel builds, but check gives this error:

    * checking package dependencies ... ERROR
    Namespace dependencies not required: ‘base64enc’ ‘repr’

IRdisplay and repr give a couple of warnings on check related to naming/documentation.  They build fine.

rmzq builds fine when I remember to `git clone --recursive`.

I have no idea how universal these builds are.  I am on OSX 10.10.2.

Where to upload?

noamross:
Files here: https://www.dropbox.com/sh/gntasgmkrpedd83/AADA_9QFMe3_8sgU2rbNA0oCa?dl=0

takluyver:
Excellent, thanks. I guess there are a couple of imports we can drop from IRkernel now.

We're just moving towards releasing these packages (#97), so if it's OK, we'll ask you to redo this process in a couple of days with the released versions, and then upload those packages to the repository.

JanSchulz:
@noamross Ok, so that works. Would you mind doing a PRs for these files? I will put my drat package somewhere and then you could use that to add your packages to the repo (see #90, which has my initial script to build the R repo: adding to a repo is basically a `R --slave -e "drat:::insert('package.tgz', repodir='irkernel.github.io')"` with the R which built the package (I use both 3.1 and 3.2)

takluyver:
Hold off on the PR until we do a release, though - I don't want to make that repo grow ridiculously by adding binaries of unreleased versions.

flying-sheep:
uh-oh, now this undoes IRkernel/IRkernel@1a392439964b9da714148f4e3799e252f43baf70. could you change this so it doesn’t?

JanSchulz:
@flying-sheep Hopefully i got everything (reduced indention, s/shell/bash/). Not sure if there is more?

takluyver:
Thanks! If this needs more fixes, we can do them in separate PRs - I'm merging now.

noamross:
Sure, just ping me when you're ready.

PR #98, opened 2015-04-24
Fix error from R CMD check

takluyver:
base64enc is no longer directly required by IRkernel, but repr is.


flying-sheep:
will we also need all individual functions so that the S3 methods work?

it doesn’t seem like it but we *do* use `repr_text` directly: IRkernel/IRkernel@9030c9a3533edad60b571965e78c7c854ce53f8f

Issue #97, opened 2015-04-24
Release? 


asmeurer:
It would be nice to get a release out with the changes from https://github.com/IRkernel/IRkernel/pull/84 so I could update the [conda package](http://binstar.org/r/r-irkernel). This will also help @rgbkrk with his work on tmpnb.

@JanSchulz’ TODO:
------------------------------

* [x] pull IRkernel/repr#3 Check capabilities before generating plot `repr`s
* [x] pull #99 Catch failures from `repr` and display the error
* [x] pull #100 and IRkernel/IRdisplay#7 display() API using `repr`
* [x] pull #102 'is_complete_request'
* [x] pull #105 Silence warning about comparing vectors
* [ ] generate Windows and Mac OSX binary packages on R 3.1 and 3.2 (cc: @noamross) and upload them to irkernel.github.io
* [ ] update Readme to mention OSX binary packages

Undecided bigger discussion issues still open: 
* [x] decide on scope of `repr` package (or postpone?) → IRkernel/repr#4
* [x] IRkernel/IRdisplay#6 and IRkernel/IRdisplay#8 Final discussion about 'IRdisplay' design 

For later:
* [x] IRkernel/IRdisplay#5 more `display_*` functions

takluyver:
@flying-sheep @JanSchulz where are we with regards to tagging a release and uploading it to the repo we've now got set up? I'm not aware of anything we particularly need to do before that.

takluyver:
Actually, we should probably change IRdisplay::display() as we discussed in IRkernel/IRdisplay#5, and then do a release.

takluyver:
aaaand we should do something about the plot display issues described in #96 . I think it's fine to just skip pdf/svg if we don't have cairo for now.

JanSchulz:
I would still argue that `repr` should only get the bare minimum of implmentations (basically only whatever can be printed via `print`. See here for my arguments: https://github.com/IRkernel/repr/commit/3ff5ba87b426ecaba0babe1d53925a6e08c52a3e#commitcomment-10888021

JanSchulz:
I also think someone with knowledge of R packaging should do a `R CMD check %G%_*.tar.gz` and clean warnings/errors :-)

JanSchulz:
I would also love to get https://github.com/IRkernel/IRkernel/issues/94 into a release

takluyver:
To borrow a phrase from Fernando: I sense a pull request in your future. ;-)

[...]

PR #99, opened 2015-04-24
Catch failures from repr and display the error

takluyver:
As mentioned in #97, now that handle_value and handle_graphics can call third-party code via repr, we should catch failures in those so they don't crash the kernel.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -65,7 +65,10 @@ execute = function(request) {
   send_plot <- function(plotobj) {
       params <- list()
       for (mime in getOption('jupyter.plot_mimetypes')) {
-          params[[mime]] <- mime2repr[[mime]](plotobj)
+          tryCatch(
+            params[[mime]] <- mime2repr[[mime]](plotobj),
+            error = handle_error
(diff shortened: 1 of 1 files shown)


flying-sheep:
good idea. should we do that or show a warning and return `NULL`?

JanSchulz:
IMO: 

* errors should be send to a kernel log
* if one or more non-text/plain repr functions work, only send the ones which return non-NULL values (so e.g. for plotting there is only one missing but plots are still displayed)


No sure what should happen, if all non-text/plain repr_* error: IMO the plain text one should be replaced by the error messages to send this to the users so that their is some info there. But if we per default send pdf and that would be the only one which is send back because all other image types error, then the notebook will only display the text plain one which will not have the error message :-(

takluyver:
In IPython, we switched from showing a warning for repr failures to just treating it as an error, so that's what I'm following here. It's easier to debug an error than something showing up not as you expect.

JanSchulz:
@takluyver ok, then LGTM

Issue #94, opened 2015-04-22
"Got unhandled msg_type:" "is_complete_request"


JanSchulz:
In knitpy I use the kernel to check if code lines are "complete" (e.g. in python the line which starts a loop is "incomplete"):

```
            msg = engine.kernel.is_complete(lines+"\n\n")
            reply = engine.kernel.get_shell_msg(timeout=self.timeout)
            assert reply['msg_type'] == 'is_complete_reply', str(reply)
            if self.kernel_debug:
                self.log.debug("completion_request: %s", msg)
            if reply['content']['status'] == 'complete':
                [...]
                # run the lines
                self._run_lines(lines+"\n", context)
                lines = ""
            elif reply['content']['status'] == 'invalid':
                # TODO: not sure how this should be handled
                # Either abort execution of the whole file or just retry with the next line?
                # However this should be handled via a user message
                self.log.info("Code invalid:\n%s",  lines)
                lines = ""
            else:
                # the "incomplete" case: don't run anything wait for the next line
                lines += "\n"
```

This doesn't work with the r kernel, as it seems that message is unhandled and the `get_shell_msg` times out with an Exception :-(

Please implement the message :-)

takluyver:
Since R is a curly braces language, if we can tokenize the code, counting `{` and `}` should be good for a first pass.

JanSchulz:
`if (....)\n{...}` might be a problem if a users submits after the closing `)`. The R repl should be able to handle it, as rstudio tells me when my line isn't finished. No idea if that info is available at all in the repl/evaluate :-(

JanSchulz:
```
> evaluate:::parse_all("if (True)")
Error in parse(text = x, srcfile = src) : 
  <text>:2:0: unexpected end of input
1: if (True)
   ^
> evaluate:::parse_all("if (True) {")
Error in parse(text = x, srcfile = src) : 
  <text>:2:0: unexpected end of input
1: if (True) {
   ^
> evaluate:::parse_all("if (True) {print('x')}")
                     src         expr
0 if (True) {print('x')} if (True....
```

takluyver:
So hopefully we can try `parse_all`, catching exceptions. If it raises an exception with the text 'unexpected end of input', it's not complete. If it's another exception, or it completes successfully, it is complete.

JanSchulz:
Another exception should probably be a message with `reply['content']['status'] == 'invalid'`:

```
> evaluate:::parse_all("if (True) {print('x)}")
[...]

PR #102, opened 2015-04-25
Implement is_complete_request

JanSchulz:
Closes: #94

This is tested against knitpy with a R kernel document, so it seems it works...

PS: @takluyver senses are good... :-)

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -1,3 +1,4 @@
+#' @importFrom evaluate parse_all
 Kernel <- setRefClass("Kernel",
                 fields = c("connection_info", "zmqctx", "sockets", "executor"),
                 methods= list(
@@ -101,9 +102,46 @@ handle_shell = function() {
         history(msg)
     } else if (msg$header$msg_type == "complete_request") {
         complete(msg)
+    } else if (msg$header$msg_type == "is_complete_request") {
+        is_complete(msg)
     } else {
         print(c("Got unhandled msg_type:", msg$header$msg_type))
     }
+},
+#'Checks whether the code in the rest is complete
+#'
+#' @param  request the is_complete request 
+#' @export
+is_complete = function(request) {
+    code <- request$content$code
+    message <- tryCatch({
+        parse_all(code)
+        # the code compiles, so we are complete (either no code at all / only
+        # comments or syntactically correct code)
+        "complete" # no function, so no return!
+    }, error = function(e)e$message)
+    # One of 'complete', 'incomplete', 'invalid', 'unknown'
+    status <- if(message == "complete"){
+        # syntactical complete code
+        "complete"
+    } else if (grepl("unexpected end of input", message)){
+        # missing closing parenthesis
+        "incomplete"
+    } else if (grepl("unexpected INCOMPLETE_STRING", message)){
+        # missing closing quotes
+        "incomplete"
+    } else {
+        # all else
+        "invalid"
+    }
+    content <- list(status = status)
+    if (status == "incomplete"){
+        # we don't try to guess the indention level and just return zero indention
+        # That's fine because R has braces... :-)
+        # TODO: do some guessing?
+        content <- c(content, indent = "")
+    }
+    send_response('is_complete_reply', request, 'shell', content)
 },
 
 complete = function(request) {


R/kernel.r
```diff
@@ -101,11 +101,51 @@ handle_shell = function() {
...
+    } else {
+        # all else
+        "invalid"
+    }
+    content = list(status=status)
+    if (status == "incomplete"){
+        # we don't try to guess the indention level and just return zero indention
+        # That's fine because R has braces... :-)
+        # TODO: do some guessing?
+        content = c(content, indent = "")
```
takluyver: I generally add things to a list by doing `content$indent = ""`. Both work, I don't know which is considered more idiomatic R.

R/kernel.r
```diff
@@ -101,11 +101,51 @@ handle_shell = function() {
...
 
+is_complete = function(request) {
+    code <- request$content$code
+    message <- tryCatch({
+        evaluate:::parse_all(code)
+        # the code compiles, so we are complete (either no code at all / only
+        # comments or syntactically correct code)
+        "complete" # no function, so no return!
+    }, warning = function(e){
+        return("warning") # IMO this can't happend...
```
takluyver: warnings aren't fatal, right? So if we don't know of a case where this can happen, maybe it's better to leave out the handler and let the warning print out.

takluyver:
This is looking good to me - @flying-sheep, can you take a quick look at it before we merge?

R/kernel.r
JanSchulz: I had to google for it and got it from here: http://stackoverflow.com/a/2436960/1380673 :-)

R/kernel.r
JanSchulz: ok, taking that out

R/kernel.r
flying-sheep: i think `c` is clearer in that you want to add a new entry instead of updating one, but since the variable is defined 2 lines above it doesn’t matter

R/kernel.r
```diff
@@ -101,11 +101,47 @@ handle_shell = function() {
...
+        is_complete(msg)
     } else {
         print(c("Got unhandled msg_type:", msg$header$msg_type))
     }
 },
 
+is_complete = function(request) {
+    code <- request$content$code
+    message <- tryCatch({
+        evaluate:::parse_all(code)
```
flying-sheep: this is [exported in `evaluate`](https://github.com/hadley/evaluate/blob/master/R/parse.r#L11), so why three “:::”? does this even work? i thought exported stuff requires exactly two “::”?

R/kernel.r
```diff
@@ -101,11 +101,47 @@ handle_shell = function() {
...
+        # the code compiles, so we are complete (either no code at all / only
+        # comments or syntactically correct code)
+        "complete" # no function, so no return!
+    }, error = function(e){
+        return(e$message)
+    })
```
flying-sheep: style: just do `tryCatch(..., error = function(e) e$message)`

R/kernel.r
```diff
@@ -101,11 +101,47 @@ handle_shell = function() {
...
+        # missing closing parenthesis
+        "incomplete"
+    } else if (grepl("unexpected INCOMPLETE_STRING", message)){
+        # missing closing quotes
+        "incomplete"
+    } else {
+        # all else
+        "invalid"
+    }
+    content = list(status=status)
```
flying-sheep: style: spaces around `=`

R/kernel.r
```diff
@@ -101,11 +101,47 @@ handle_shell = function() {
...
+    message <- tryCatch({
+        evaluate:::parse_all(code)
+        # the code compiles, so we are complete (either no code at all / only
+        # comments or syntactically correct code)
+        "complete" # no function, so no return!
+    }, error = function(e){
+        return(e$message)
+    })
+    # One of 'complete', 'incomplete', 'invalid', 'unknown'
+    status <- if(message=="complete"){
```
flying-sheep: style: spaces around `==`

R/kernel.r
JanSchulz: done, hopefully :-)

R/kernel.r

R/kernel.r

R/kernel.r
JanSchulz: done, also '<-' instead of '='

JanSchulz:
pushed, but untested (didn't install due to jsonline and have to run :-( )

takluyver:
Thanks, I think that looks good. I'll merge and run roxygenise to update NAMESPACE

PR #105, opened 2015-04-25
Silence warning about comparing vectors

takluyver:
@JanSchulz - I think this should get rid of the remaining warning you saw.


JanSchulz:
Yes, tried this branch and shutdown didn't generate any warning anymore. -> LGTM

PR #106, opened 2015-04-26
Switch to HTTPS on github.io

rgbkrk:
https://irkernel.github.io/ totally works

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ We provide windows binary packages of all the needed packages. The packages
 and the kernel spec can be installed with the following lines in an R console: 
 
 ```r
-install.packages(c("rzmq","repr","IRkernel","IRdisplay"), repos="http://irkernel.github.io/")
+install.packages(c("rzmq","repr","IRkernel","IRdisplay"), repos="https://irkernel.github.io/")
 IRkernel::installspec()
 ```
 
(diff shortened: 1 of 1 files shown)


takluyver:
Good catch

takluyver:
@JanSchulz do we need to specify the dependencies in `install.packages()`? I'd expect it to use the same repo when installing dependencies, in which case we should just be able to recommend:

```R
install.packages("IRkernel", repos="https://irkernel.github.io/")
```

JanSchulz:
@takluyver AFAIK we don't need dependencies in `install.packages`, but I used them to install everything even if a package with the same version was already installed.

We might also want to switch to drat (see here for a comment: https://github.com/eddelbuettel/drat/pull/16#issuecomment-96288353) , which would get us this:

```
drat::addRepo("irkernel")           # short, sweet, simple -> add to ~/.Rprofile
install.packages("IRkernel")        # assuming the other are Depends
update.packages()                   # no need for any repos argument
```

* + short, kind of an evolving standard
* + adding it to Rprofile gets `update.packages` for free 
* ~ `update.packages` can also be achieved via three lines in  `~/.Rprofile`
* - one more package to install as a user
* - installation without drat will need a slightly different URL (irkernel.github.io/drat)

takluyver:
Let's leave it as is for now - `drat` may pay off later, but for initial installs it's a couple of extra steps. I like that with the current setup a single command can get the packages installed.

PR #110, opened 2015-04-28
Start adding tests using new kerneltest tool

takluyver:
This isn't ready for merge yet, but I wanted to show what we're working on. This will start up the installed IRkernel and test it by sending and receiving messages.

The kerneltest tool is being developed at https://github.com/jupyter/kerneltest

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+language: python
+python:
+  - "3.5"
+install:
+  - sudo apt-get update
+  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
+  - bash miniconda.sh -b -p $HOME/miniconda
+  - export PATH="$HOME/miniconda/bin:$PATH"
+  - hash -r
+  - conda config --set always_yes yes --set changeps1 no
+  - conda update -q conda
+  # Useful for debugging any issues with conda
+  - conda info -a
+
+  # Replace dep1 dep2 ... with your dependencies
+  - conda create -q -n test-environment -c r r r-base64enc r-evaluate r-jsonlite r-irdisplay r-repr jupyter pip
+  - source activate test-environment
+  - pip install jupyter_kernel_test
+  - R CMD INSTALL $TRAVIS_BUILD_DIR
+  - R -e "IRkernel::installspec()"
+
+script:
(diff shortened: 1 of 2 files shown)


flying-sheep:
we could integrate this into whatever R testing framework we choose to use by querying a [TestLoader](https://docs.python.org/3.4/library/unittest.html#unittest.TestLoader)

flying-sheep:
ugh, i tried to hack on this but getting all this stuff installed without it being on cheeseshop suuucks.

after getting to “traitlets.config: there is no such package”, i gave up.

takluyver:
Yes, sorry, for the moment it depends on a development version of traitlets, along with several other packages. I think we should be releasing traitlets in the next few days, which will make it a bit easier to get started. I was writing this along with `jupyter_kernel_test` to validate the machinery - it's not yet in a state where I'd ask other people to install it.

takluyver:
All the pieces for this are now released: you should be able to `pip install jupyter_kernel_test` and run `python test_ir.py` to run the tests. I'll start trying to work out how we can run the tests on Travis.

takluyver:
Tada, Travis tests! :sparkles:

flying-sheep:
yay!

takluyver:
:-) I had been putting this off because I wasn't sure how to run tests that need both Python and R on Travis. Then I remembered the option of setting it up with conda, and it became pretty straightforward.

flying-sheep:
now we just have to add tests for everything :wink:

takluyver:
I certainly hope we can get some conventional R tests on repr. But it's comforting to know that we'll found out if we completely break the IRkernel code!

flying-sheep:
jup. we only still need to think about testing plot reprs there

JanSchulz:
If oyu anyway use conda, then install matplotlib, they have some functions which can be used to compare two images. ggplot uses that to implement a assert for that: https://github.com/yhat/ggplot/blob/master/ggplot/tests/__init__.py#L80:

the main point in that function is (after writing the current image to disk and cmputing the filename to the baseline image):

```
    err = compare_images(expected_fname, actual_fname,
                         tol, in_decorator=True)
    if err:
        msg = 'images not close: {actual:s} vs. {expected:s} (RMS {rms:.2f})'.format(**err)
        raise ImagesComparisonFailure(msg)
```

PR #115, opened 2015-04-29
Less printing on startup

takluyver:
This removes the printing of the connection info, and suppresses the default R banner.

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -219,7 +219,6 @@ shutdown = function(request) {
 
 initialize = function(connection_file) {
     connection_info <<- fromJSON(connection_file)
-    print(connection_info)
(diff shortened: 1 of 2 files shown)


R/kernel.r
```diff
@@ -219,7 +219,6 @@ shutdown = function(request) {
 
 initialize = function(connection_file) {
     connection_info <<- fromJSON(connection_file)
-    print(connection_info)
```
JanSchulz: I think we need a debug frag, so that these info can be seen by some parameter...

takluyver: Jupyter/IPython doesn't currently pass any such flags to kernels, though. So for now, let's just edit in debugging prints where we need them. It's been a long time since I needed to refer to this information for debugging, anyway.

rgbkrk:
Yay, thank you for this.

PR #116, opened 2015-04-30
http instead of https

randy3k:
as title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ We provide Windows and Mac OS X binary packages of all the needed packages. The 
 and the kernel spec can be installed with the following lines in an R console: 
 
 ```r
-install.packages(c('rzmq','repr','IRkernel','IRdisplay'), repos = 'https://irkernel.github.io/')
+install.packages(c('rzmq','repr','IRkernel','IRdisplay'), repos = 'http://irkernel.github.io/')
 IRkernel::installspec()
 ```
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
What's preventing you from using https?

randy3k:
I don't know, it just doesn't work

```
> install.packages(c('rzmq','repr','IRkernel','IRdisplay'), repos = 'https://irkernel.github.io/', type = 'source')
Installing packages into ‘/Users/Randy/Library/R/3.2/library’
(as ‘lib’ is unspecified)
Warning: unable to access index for repository https://irkernel.github.io/src/contrib
Warning message:
packages ‘rzmq’, ‘repr’, ‘IRkernel’, ‘IRdisplay’ are not available (for R version 3.2.0) 
```

It works if i use http.

randy3k:
The R function `download.file` doesn't support https by default. https://stat.ethz.ch/R-manual/R-devel/library/utils/html/download.file.html

rgbkrk:
Now that's ugly. :unamused: 

I'm so sorry. I have libcurl setup on all my systems, but I know that I'm probably an outlier (Linux or OS X only), since I'm assuming most R users are on Windows. :fallen_leaf:

takluyver:
Just spoken to @jeroenooms about this. Supplying `method='curl'` makes it work if curl is installed. But I imagine it won't be available on Windows (with any reliability). 'libcurl' is a new feature in R 3.2.0 - I'm not sure if it will be there on Windows or not.

It seems like a pretty serious security hole that packages are installed from http by default, but I don't know what else we can do for this.

randy3k:
Perhaps rolling back to the `devtools` approach?

takluyver:
Perhaps, but I'd like to avoid that - it's significantly more steps for
users. Can devtools even install binary packages? If not, it's not really
an option. And if devtools itself has to be installed over http, it's not
really a security benefit.

jeroenooms:
Best to just stick with http for now, like all other packages. This problem will eventually solve itself with better native ssl support in R coming up.

randy3k:
@jeroenooms :+1:

rgbkrk:
Great to hear @jeroenooms! :sweat_smile:

takluyver:
Thanks all

rgbkrk:
Thank you @randy3k, sorry for the trouble.

PR #133, opened 2015-05-17
Don't send rich text response when there is no console output

abielr:
When generating the rich text response to display an R object, first
check that there is console output. If the object would not be
displayed at the console then do not generate a response, so that the
notebook more closely mimics what would be displayed at the console.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -98,21 +98,25 @@ execute = function(request) {
   } else {
     handle_value <- function (obj) {
         data <- namedlist()
-        if (getOption('jupyter.rich_display')) {
-            tryCatch(
-                for (mime in getOption('jupyter.display_mimetypes')) {
-                    r <- mime2repr[[mime]](obj)
-                    if (!is.null(r)) data[[mime]] <- r
-                },
-                error = handle_error
-            )
-        } else {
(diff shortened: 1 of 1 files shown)


flying-sheep:
[looks good!](https://github.com/IRkernel/IRkernel/pull/133/files?w=1) thanks!

R/execution.r
```diff
@@ -98,21 +98,25 @@ execute = function(request) {
...
-            data = data,
-            metadata = namedlist(),
-            execution_count = execution_count))
+        data[['text/plain']] <- repr_text(obj)
+        
+        # Only send a response when there is regular console output
+        if (nchar(data[['text/plain']]) > 0) {
+            if (getOption('jupyter.rich_display')) {
+                tryCatch(
+                    for (mime in getOption('jupyter.display_mimetypes')) {
```
takluyver: I think this will now calculate the text/plain repr twice if that mimetype is in `jupyter.display_mimetypes`. I can't imagine the delay will be noticeable, though.

flying-sheep: noticed that as well and thought the same.

we could do a `setdiff` to remove `text/plain` from the mimetypes to fix it if we want.

Issue #134, opened 2015-05-17
Is there a workable installation instruction on OS X?


ywhuofu:
Thanks for this package. I very much look forward to use it, but the instruction simply doesn't work. Could anyone who managed to get this work post a valid installation instruction please? 

Thanks !

takluyver:
What do you see when you follow the instructions at http://irkernel.github.io/ ? It's hard to resolve problems when you just tell us something "doesn't work"!

ywhuofu:
Sorry, I should be specific. The error was "fail to compile zmq", a long standing issue I guess. I looked for various solutions online but none of them worked.

takluyver:
The instructions there for OS X should install a binary package, so it shouldn't need to compile zmq. Can you copy the full output from what you did, including the commands you ran into a gist.

Ping @JanSchulz - I know there was some change to the repo layout recently, could it be failing to find binary packages and falling back on source?

ywhuofu:
Please see below, the commands I pasted into R terminal is straight from the instruction `install.packages(c('rzmq','repr','IRkernel','IRdisplay'), repos = 'http://irkernel.github.io/')`

### Below is my terminal output
---------------------
```
R version 3.2.0 (2015-04-16) -- "Full of Ingredients"
Copyright (C) 2015 The R Foundation for Statistical Computing
Platform: x86_64-apple-darwin13.4.0 (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> install.packages(c('rzmq','repr','IRkernel','IRdisplay'), repos = 'http://irkernel.github.io/')
Warning: dependencies ‘uuid’, ‘base64enc’ are not available
Warning: unable to access index for repository http://irkernel.github.io/bin/macosx/mavericks/contrib/3.2
Package which is only available in source form, and may need
  compilation of C/C++/Fortran: ‘rzmq’
Do you want to attempt to install these from sources?
y/n: y
installing the source packages ‘rzmq’, ‘repr’, ‘IRkernel’, ‘IRdisplay’

[...]

PR #135, opened 2015-05-19
Suggest installing the xcode client tools

rgbkrk:
When trying to figure out https://github.com/IRkernel/IRkernel/issues/134, even with zmq and czmq installed I was truly scratching my head.

Turns out that clang can't find those headers by default until you install the xcode command line developer tools.

```
xcode-select --install
```

Oddly, with the binary packages that are up on irkernel.github.io, I still needed to have run `xcode-select --install`...

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ You'll need zmq development headers to compile rzmq.
 *   **Homebrew**
 
     ```bash
+    xcode-select --install
     brew install zmq
     # or upgrade
     brew update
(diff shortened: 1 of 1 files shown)


rgbkrk:
Also, yes. clang. Even though I have gcc on the machine, apparently R aims for native and chooses clang for `CC`.

takluyver:
Thanks, sorry I overlooked this for a while.

Issue #145, opened 2015-06-01
Rendering of graphs in R kernel Notebooks on GitHub


cfljam:
Hi
I have been making some example notebooks about use of the kernel and R2Py for R authoring in Jupyter but have struck some inconsistencies in notebook rendering on GitHub. 

Whereas my python kernel notebook using R2py displays fine https://github.com/cfljam/pyRat/blob/master/Notebooks/Using%20R2Py%20in%20Jupyter%20NOtebooks.ipynb

but the same code using IRKernel doesnt display them https://github.com/cfljam/pyRat/blob/master/Notebooks/Jupyter%20Notebooks%20with%20the%20R%20kernel.ipynb

My apologies  if this is off-topic and should direct this to Github but I would be grateful of any suggestions. The automatic rendering of Jupyter notebooks in GitHub is a huge attraction and so a major motivation for using this kernel.

rgbkrk:
Interesting. It doesn't render on nbviewer either. http://nbviewer.ipython.org/github/cfljam/pyRat/blob/master/Notebooks/Jupyter%20Notebooks%20with%20the%20R%20kernel.ipynb

I do see the plots in the raw notebook (5 different display options!!!).

takluyver:
Sounds like jupyter/nbconvert#5. Kyle, do you know what version of nbconvert is running on nbviewer (& Github's renderer)?

cfljam:
I have also struck some similar cases of missing textual output from %%bash cell magics when rendered with nbviewer. I will try to find a reproducible example of this.

Many thanks for looking into this...predictable rendering of content would make the R kernel+ Jupyter  a **very** attractive  alternative to Rmd.

rgbkrk:
Always latest release since last merged PR/deploy, which appears to be 5 days ago (via breadcrumbs at the bottom of any nbviewer page).

cfljam:
Exactly the same result (ie missing images) rendered with
````
root@940e5f39d03b:/Users/johnmccallum/Documents/pyRat/Notebooks# ipython nbconvert --version
3.2.0-dev
```

flying-sheep:
well, the fix was in *Jupyter* nbconvert. IPython nbconvert still has the bug.

the 5 display options are overkill, but the amount of bugs we discovered in several IPython/Jupyter components by leaving it in is worth it imho. (although we could remove JPEG, which leaves PNG, SVG, PDF, and the dummy text output)

takluyver:
[...]

PR #154, opened 2015-06-17
Remove JPEG and PDF from default plot output formats

takluyver:
Leaving png, svg and text.

From discussion on #145


rgbkrk:
:+1: Looks good.

flying-sheep:
yeah, let’s do this!

PR #155, opened 2015-06-19
Cleanup

flying-sheep:
Cleaned up everything.

Since this includes a few logic changes, please look at it before we merge.

and don’t start new work and branches not based on this, as everything will conflict.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -18,9 +18,11 @@ plot_builds_upon <- function(prev, current) {
     if (is.null(prev)) {
         return(TRUE)
     }
+    
     lprev <- length(prev[[1]])
     lcurrent <- length(current[[1]])
-    return((lcurrent >= lprev) && (identical(current[[1]][1:lprev], prev[[1]][1:lprev])))
+    
+    lcurrent >= lprev && identical(current[[1]][1:lprev], prev[[1]][1:lprev])
 }
 
 Executor <- setRefClass(
--- a/R/execution.r
+++ b/R/execution.r
@@ -1,6 +1,8 @@
 #' @include options.R
 NULL
 
+setClassUnion('recordedplotOrNULL', members = c('recordedplot', 'NULL'))
+
 displayenv <- environment(publish_mimebundle)
 
 lappend <- function(lst, obj) {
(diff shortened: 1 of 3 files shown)


R/kernel.r
```diff
@@ -169,94 +192,100 @@ complete = function(request) {
...
-                        file_extension=".r", version=rversion
-                     ),
-                     banner=version$version.string
-                    )
-                )
+    rversion <- paste0(version$major, '.', version$minor)
+    send_response('kernel_info_reply', request, 'shell', list(
+        protocol_version       = '5.0',
+        implementation         = 'IRkernel',
+        implementation_version = '0.2',
```
JanSchulz: should this be 0.4?

takluyver: We should probably define a global version variable, so that it's easier to remember to update it.

R/execution.r
```diff
@@ -1,12 +1,14 @@
 #' @include options.R
 NULL
 
-displayenv = environment(publish_mimebundle)
+setClassUnion('recordedplotOrNULL', members = c('recordedplot', 'NULL'))
```
takluyver: @flying-sheep Can I suggest that functional changes to the code go in a different PR from the reformatting and style fixes? It's easier to review actual changes if they're not scattered through a big diff of trivial changes.

flying-sheep:
Of course. That one was a mistake to change yet

flying-sheep:
so can we pull this in?

flying-sheep:
hey guys!

@takluyver 

@JanSchulz 

it would be very nice to use this as a base going on :smile:

flying-sheep:
well, then i’ll simply do it :smile:

R/kernel.r
flying-sheep: using `packageVersion` now

R/execution.r
flying-sheep: Of course. That one was a mistake to change yet

takluyver:
Sorry, I forgot about this. I think it's all fine, though I'd still have preferred the functional change I commented on before to be a separate PR.

flying-sheep:
ok, as said: it was a mistake, but worked quite well for some weeks for me.

PR #172, opened 2015-07-29
Only pass response function to executor

flying-sheep:
This way there is no two-way dependency:

The executor only knows how to send a response, not more.


takluyver:
Seems nice, thanks.

Issue #120, opened 2015-05-01
Can't initiate IRKernel in Windows 10


MengyingLi:
Hi I am currently using Windows 10 and followed the instruction about using R to install IRKernel 
![image](https://cloud.githubusercontent.com/assets/6514708/7433312/d5083c4e-efe4-11e4-94f8-3ccfb3069fb2.png)

Everything seems to be fine but there is no R kernel in the drop down menu from Jupyter notebook.  did I miss anything? 
![image](https://cloud.githubusercontent.com/assets/6514708/7433323/fba8ad20-efe4-11e4-9e3a-5f5b29c407f5.png)

takluyver:
That looks like it should have worked. Can you look in `~/.ipython/kernels` and see if there's an `ir` directory there?

MengyingLi:
Hi! Thanks for the prompt reply.  I did see an ir directory there. I installed python using anaconda. Is that because the directory are different?  
![image](https://cloud.githubusercontent.com/assets/6514708/7443750/84f1b260-f117-11e4-91a0-8262439a20b4.png)

takluyver:
That should work with anaconda. Apologies if I'm checking the obvious, but
did you reload the page after you installed IRkernel?

MengyingLi:
Hi, Thank you so much for your help! :) I actually installed python 2.7 first and then add a python 3 to anaconda. Here is the script I initiated python 3 
![image](https://cloud.githubusercontent.com/assets/6514708/7445795/a31d6b12-f176-11e4-81b7-78fc6f03d010.png)

I also checked the ipython folder under the anaconda directory and there is no IR kernel in the kernel folder

![image](https://cloud.githubusercontent.com/assets/6514708/7445801/d45b6a9e-f176-11e4-8ae9-3eb8db8e1cd6.png)

takluyver:
Strange. Can you run this and post the output:

```python
from IPython.kernel.kernelspec import KernelSpecManager
ksm = KernelSpecManager()
ksm.kernel_dirs
```

And also:

```python
ksm.find_kernel_specs()
```

MengyingLi:
Hi Attached is the output from the commands:

![image](https://cloud.githubusercontent.com/assets/6514708/7456420/46a706cc-f239-11e4-9c51-849cadb160af.png)

takluyver:
[...]

PR #173, opened 2015-07-30
installspec: use full path to R interpreter

dougalsutherland:
This patch changes `installspec` to use the full path to the current R interpreter called by the process, rather than just using whatever `R` comes first on `ipython`'s `PATH`.

This makes life easier when R isn't on your path by default (as in #120 and everything mentioned by #161), or when you have multiple R installs and whatever's launching your ipython notebook server defaults to the wrong one (as in my case where I was pretty confused just now). It also matches the behavior of e.g. IJulia or `ipython kernelspec install-self`.

Not tested on Windows; someone should confirm it works there before merging.

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -323,8 +323,19 @@ installspec <- function(user = TRUE) {
     if (!found_binary)
         stop('IPython has to be installed but could neither run ipython nor ipython2 or ipython3.')
     
+    # make a kernelspec with the current interpreter's absolute path
     srcdir <- system.file('kernelspec', package = 'IRkernel')
+    tmp_name <- tempfile()
+    dir.create(tmp_name)
+    file.copy(srcdir, tmp_name, recursive = TRUE)
+    spec_path <- file.path(tmp_name, 'kernelspec', 'kernel.json')
+    spec <- fromJSON(spec_path)
+    spec$argv[[1]] <- file.path(R.home('bin'), 'R')
+    write(toJSON(spec, pretty = TRUE), file = spec_path)
+
     user_flag <- if (user) '--user' else character(0)
-    args <- c('kernelspec', 'install', '--replace', '--name', 'ir', user_flag, srcdir)
+    args <- c('kernelspec', 'install', '--replace', '--name', 'ir', user_flag, file.path(tmp_name, 'kernelspec'))
     system2(binary, args, wait = TRUE)
+
+    unlink(tmp_name, recursive = TRUE)
 }
--- a/README.md
+++ b/README.md
@@ -14,9 +14,6 @@ To run IPython with an R kernel, you need at least the following:
   way to get everything you need. Earlier versions of IPython are now 
   unsupported.
 * A current [R installation](http://www.r-project.org/).
-* You must be able to run `R` at a command prompt to start R. On Windows, you will
-  likely need to [edit the system path](http://www.computerhope.com/issues/ch000549.htm)
-  to add the directory containing `R.exe`.
 
 Installing from source needs additionally header files (see below).
 


takluyver:
Thanks. Hopefully someone with Windows can test it out soon.

@bigzachattack will be happy about this, while @asmeurer may need to work out how it affects conda packaging.

dougalsutherland:
The conda fix is probably just involves putting `kernel.json` into `info/has_prefix` in the package, but @asmeurer knows more about that than I do.

bigzachattack:
👍

asmeurer:
+100 for using the install R's location rather than searching the PATH. For the conda package, we [install the kernspec manually](https://github.com/conda/conda-recipes/blob/master/r-packages/r-irkernel/bld.bat) (I think because the default is to install to the user space, which is completely wrong).

asmeurer:
I don't know much R, but what is the tempfile stuff doing?

takluyver:
Copy the kernelspec directory to a temporary directory, modify kernel.json, and then invoke ipython kernelspec install with the temporary directory.

dougalsutherland:
I don't know much about R either; that seemed like the easiest way to get it in place without hardcoding in any of `ipython kernelspec`'s logic. If there's a simpler suggestion I'd be happy to hear it.

flying-sheep:
hi!

so first, please add a check to not do that on linux. on linux, the PATH works, so no need to circumvent it.

second, you can use `tempdir()` instead of defining a `tmp_name`, as `tempdir()` returns a per-session temporary directory. i.e. remove the lines `tmp_name <- tempfile()` and `dir.create(tmp_name)`, and use `tempdir()` instead of `tmp_name`.

and last (sorry for being anal) to conform with coding style, please replace `"` with `'` and `spec$argv[1]` with `spec$argv[[1]]`

bigzachattack:
If you have multiple envs in anaconda and you want to access R from one env and IPython is in the other, then the PATH does not work because Anaconda will disregard one path for the other.

flying-sheep:
please elaborate, i’m not familiar with anaconda envs

bigzachattack:
For instance you could have two anaconda environmnets with the following paths:
```
/opt/conda/envs/python3
/opt/conda/envs/r
```
Each env has its own `bin/`, `lib/`, etc.    

To switch into an env:
```
source activate <env>
```

This will change the profile by disregard any existing Anaconda env in PATH and replacing it with `/opt/conda/envs/<env>/bin`.  By default `/opt/conda/bin` is in your path.  

When you run `source activate r`
> `/opt/conda/bin` is replaced with `/opt/conda/envs/r/bin` in PATH. 

Now R would be in your path.

If you then `source activate python3`
> `/opt/conda/envs/r/bin` is replaced with `/opt/conda/envs/python3/bin`

Now R would not be in your path.


Probably the majority of people will install R in a place where it is in PATH: either system-wide or when using Anaconda install irkernel and ipython-notebook in the same env.  However, there will be a set of  people that have non-standard installs where they cannot have R in the PATH with jumping through extra hoops.

dougalsutherland:
Yes, having multiple installs of R is just as much a problem on Linux as anywhere else. If you want I could add an option for whether to make the kernelspec with absolute path or not.

I did the subdirectory just to avoid a possible clash with someone else in the R session making a temp directory named "kernelspec", since it seems copy.file doesn't let you change the name of the target directory. This isn't particularly likely, but given that the cost is one line of code and like two milliseconds in a function called once per install, I thought is that was worth avoiding the possibility.

I'll make the style changes later today.

dougalsutherland:
Style changes made.

flying-sheep:
ok, i’m convinced that there are use cases on non-windows. it’s pretty advanced though, so i trust that people with workflows like this know what they want and would be able to chenge the default.

therefore, could you please add the option with a default value like this? `absolute.path = Sys.info()$sysname != 'Linux'`

takluyver:
I don't think the default should be platform dependent. The ipython kernel
already encodes the full path to the interpreter at install time, on all
OSs. That should be fine for the default here, too.

flying-sheep:
> The ipython kernel already encodes the full path to the interpreter at install time

what do you mean with “encode” here?

takluyver:
Sorry, I was typing on a touchscreen, which never lends itself to writing much detail. I meant that, when you do `ipython kernelspec install-self`, it does something similar to what's proposed here: it gets the full path of the Python interpreter using `sys.executable`, and embeds that in the argv written in kernel.json. I don't think we even have an option to make a kernelspec that just invokes `python` from $PATH. So I think it's fine for IRkernel to do the same by default when installing its kernelspec.

flying-sheep:
hmm, makes sense!

dougalsutherland:
Just a note that someone should still verify this works on Windows. I expect it to, but don't have an R install on any Windows machines to test....

PR #196, opened 2015-10-05
added quit functions that work

flying-sheep:
sans history.

is this a good way to do it? or should we rather put the quit functions elsewhere?

we need to think about how to do history: should we use R’s internal machinery somehow? or roll our own? we need to implement `history_request`.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -25,6 +25,18 @@ plot_builds_upon <- function(prev, current) {
     lcurrent >= lprev && identical(current[[1]][1:lprev], prev[[1]][1:lprev])
 }
 
+ask <- function(prompt = '') {
+    answer <- NA
+    while (is.na(answer)) {
+        answer <- switch(readline(prompt),
+           y = , Y = TRUE,
+           n = , N = FALSE,
+           c = NULL,
+           NA)
+    }
+    answer
+}
+
 Executor <- setRefClass(
     'Executor',
     fields = list(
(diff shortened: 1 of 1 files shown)


R/execution.r
```diff
@@ -25,6 +25,18 @@ plot_builds_upon <- function(prev, current) {
     lcurrent >= lprev && identical(current[[1]][1:lprev], prev[[1]][1:lprev])
 }
 
+ask <- function(prompt = '') {
+    answer <- NA
+    while (is.na(answer)) {
+        answer <- switch(readline(prompt),
```
takluyver: This is terminal interaction, right? I think we should avoid anything that does that until we've implemented the stdin part of the protocol. Otherwise it will lead to bizarre hangs when you try to shut it down.

R/execution.r
```diff
@@ -75,6 +87,26 @@ execute = function(request) {
         payload <<- lappend(payload, list(source = 'page', data = mimebundle))
     })
     
+    # .Last doesn’t seem to work, so replicating behavior
+    quit <- function(save = 'default', status = 0, runLast = TRUE) {
+        save = switch(save,
+            default = , yes = TRUE,
```
takluyver: What does this `default = ` syntax without a value do? :-S

flying-sheep: fallback to next case. basically this is

```js
var quit = switch(save) {
	'default':
	'yes':
		true
	'no':
		false
	...
}
```

R/execution.r
flying-sheep: yeah, `quit()` mainly makes sense in the context of `jupyter (qt)console --kernel=ir`

Issue #215, opened 2015-11-11
Can't input multibyte character within IRkernel of jupyter notebook?


zggjtsgzczh:
I have win7. When I used jupyter notebook with R kernel running, I can't use chinese character within R code because of dead kernel. It gives me error message on terminal: 
***
[InstallKernelSpec] Installed kernelspec ir in C:\ProgramData\jupyter\kernels\ir
: lexical error: invalid char in json text.
***
And then, Kernel Restarting. The kernel appears to have died. It will restart automatically.
Can it be resolve? thanks

my code in IRkernel of jupyter notebook:
lsearching <- "行政法"
grep("行政法",lsearching)
***
R 3.1.1
win7
python 2.7
jupyter notebook

flying-sheep:
can you give us the console output of the kernel?

zggjtsgzczh:
To flying-sheep:
console output of the kernel error message:

[I 15:26:36.549 NotebookApp] Serving notebooks from local directory: E:\temp\pyt
hon_temp\
[I 15:26:36.549 NotebookApp] 0 active kernels
[I 15:26:36.549 NotebookApp] The IPython Notebook is running at: http://localhos
t:8888/
[I 15:26:36.552 NotebookApp] Use Control-C to stop this server and shut down all
 kernels (twice to skip confirmation).
[I 15:26:55.016 NotebookApp] Kernel started: 6c9a8a8b-5273-444e-a051-645f51f8e67
f
[1] "Got unhandled msg_type:" "comm_open"
[InstallKernelSpec] Removing existing kernelspec in C:\ProgramData\jupyter\kerne
ls\ir
[InstallKernelSpec] Installed kernelspec ir in C:\ProgramData\jupyter\kernels\ir

: lexical error: invalid char in json text.
          arching <- \""\ngrep(\"",lsear
                     (right here) ------^

[I 15:27:10.017 NotebookApp] KernelRestarter: restarting kernel (1/5)
WARNING:root:kernel 6c9a8a8b-5273-444e-a051-645f51f8e67f restarted
[I 15:28:54.844 NotebookApp] Saving file at /Untitled1.ipynb

flying-sheep:
the output seems to come from [YAJL](https://github.com/lloyd/yajl)

@lloyd any idea why this might happen?

flying-sheep:
oh, and the fact that this already happens when installing the kernelspec shows that there’s probably something wrong with your installation of https://github.com/jeroenooms/jsonlite

[...]

PR #224, opened 2015-12-02
Mark JSON message data as utf-8 before parsing

takluyver:
Closes gh-215

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -40,15 +40,28 @@ wire_to_msg = function(parts) {
         expected_signature <- sign_msg(parts[(i + 2):(i + 5)])
         stopifnot(identical(signature, expected_signature))
     }
-    header        <- fromJSON(rawToChar(parts[[i + 2]]))
-    parent_header <- fromJSON(rawToChar(parts[[i + 3]]))
-    metadata      <- fromJSON(rawToChar(parts[[i + 4]]))
-    content       <- fromJSON(rawToChar(parts[[i + 5]]))
+
+    # Convert the four key parts of the message to strings and parse the JSON
+    s <- rawToChar(parts[[i + 2]])
+    Encoding(s) <- "UTF-8"
+    header      <- fromJSON(s)
+    s <- rawToChar(parts[[i + 3]])
+    Encoding(s) <- "UTF-8"
+    parent_header <- fromJSON(s)
+    s <- rawToChar(parts[[i + 4]])
+    Encoding(s) <- "UTF-8"
+    metadata    <- fromJSON(s)
+    s <- rawToChar(parts[[i + 5]])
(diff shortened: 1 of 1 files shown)


takluyver:
@flying-sheep I ran into another unicode issue while testing this. If R thinks it can't display a character, it escapes it like this: `<U+884C>` (vs Python style `\u884c`). These sequences are being included raw in the HTML repr produces, so the browser tries to interpret them as HTML tags and doesn't show anything. repr should probably be escaping strings for the HTML representation.

flying-sheep:
honestly that’s a jsonlite bug. but no harm in explicitly setting the encoding before they fix their stuff

jeroenooms:
@flying-sheep no it's not. In R each string is marked with an encoding. If you are creating a character vector based on a `const char` in utf8 encoding but you tell R that it should treat it as native encoding, then the bug is on you. Almost all string methods will fail obviously give strange results on this object. 

The `jsonlite` package will try convert strings into utf8, which obviously leads to strange results if you had marked the string incorrectly in the first place.

flying-sheep:
jup, i realized that seconds before you posted :laughing: 

but jsonlite has a bug: `toJSON` doesn’t declare an encoding!

takluyver:
I suspect it's just ambiguous. JSON is defined in terms of characters (well, unicode code points), with a note that the default encoding is utf-8. It looks like R has a single type for strings and bytes, like Python 2, so you can argue either that jsonlite is getting characters and should treat them as whatever characters R thinks they are, or that jsonlite is getting bytes and should treat them as UTF-8.

Either way, there's nothing wrong with this, and it makes stuff work. :-)

flying-sheep:
jup, edited my comment :smile: 

i also DRYed your code in 64d069474f43b9bfdee7d04e58e080a160502972

jeroenooms:
@takluyver  if `jsonlite::fromJSON` would always assume utf8, windows users would get very strange behaviour. For example they could not even copy-paste something like this in the console:

```r
fromJSON('{"food" : "寿司"}')
```

Because the console is native encoding, so `fromJSON` would actually have to treat this string as something else than what is actually displayed? 

Really there is only one proper solution here, you must mark your strings with the correct encoding :-) Once you do, R and jsonlite will automatically convert between encodings when needed.

flying-sheep:
jup, it’s done. as said, i realized that jsonlite does the right thing!

i wish everything could be utf-8 already :angry:

takluyver:
Well, that's actually up to R's Windows interface - AIUI it's possible to read UTF-16 from the Windows console, which could then be translated unambiguously to UTF-8. But most Unix-y programs (including Python, for now, and presumably R too) ignore that, and use the older interfaces that give them bytes in the current 'code page'.

Issue #245, opened 2016-01-25
interrupting the kernel while polling (inactive) kills it


Anne623:
Hi ,All

   when I click interrupt button in notebook page, the r kernel will be dead when there are no cell is running..it will show dialog:

```
Kernel Restarting
The kernel appears to have died. It will restart automatically.
```

but if interrupt a running cell,it works well.

flying-sheep:
is this the git version or a release?

takluyver:
I could replicate it - #246 fixes it for me.

flying-sheep:
ah, it occurs in both pbdZMQ and rzmq versions.

Anne623:
Hi, it not working, i use release V0.5, and i add try catch in 
```
 while (TRUE) {
            tryCatch({
                events <- poll.socket(
                    list(sockets$hb, sockets$shell, sockets$control),
                    list('read', 'read', 'read'), timeout = -1L)
                if (events[[1]]$read) {
                    # heartbeat
                    hb_reply()
                }
                if (events[[2]]$read) {
                    # Shell socket
                    handle_shell()
                }
                if (events[[3]]$read) {  # Control socket
                    handle_control()
                }
            }, interrupt=function(cond) {}
            )
        }
```

it also will be kill kernel:
```
Error in poll.socket(list(sockets$hb, sockets$shell, sockets$control),  : 
  Interrupted system call
Calls: <Anonymous> ... tryCatchList -> tryCatchOne -> doTryCatch -> poll.socket -> .Call
Execution halted
```

Anne623:
@flying-sheep yes,it's release V0.5

flying-sheep:
yes, we’re currently discussing this in #246

flying-sheep:
Fixed in 771c6fb5309d6bd0fc162a73058ebc59dda46389

PR #246, opened 2016-01-25
Catch interrupts in main loop

takluyver:
This stops the kernel from dying if it gets an interrupt outside executing user code.

Closes gh-245

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -270,6 +270,7 @@ initialize = function(connection_file) {
 
 run = function() {
     while (TRUE) {
+      tryCatch({
         zmq.poll(
             c(sockets$hb, sockets$shell, sockets$control),
             rep(.pbd_env$ZMQ.PO$POLLIN, 3))
@@ -282,7 +283,8 @@ run = function() {
         
         if(bitwAnd(zmq.poll.get.revents(3), .pbd_env$ZMQ.PO$POLLIN))
             handle_control()
-        
+      }, interrupt=function(cond) {}
+      )
     }
 })
 )


flying-sheep:
doesn’t work for me. i still get:

```
[I 11:58:54.817 NotebookApp] Kernel interrupted: 85f81956-373d-4cb3-8c60-4cf05bed6d30

Warnmeldung:
In zmq.poll(c(sockets$hb, sockets$shell, sockets$control), rep(.pbd_env$ZMQ.PO$POLLIN,  :
  R_zmq_poll: 4 strerror: Unterbrechung während des Betriebssystemaufrufs
Ausführung angehalten
[I 11:58:55.812 NotebookApp] KernelRestarter: restarting kernel (1/5)
```

i tried more stuff but nothing seems to work.

and two questions:

1. why did you do the handle_control thing?
2. could you please add me as owner to the IRkernel team? i’d like to create labelsfor bugs, change urls and descriptions, …

takluyver:
Are you definitely installing it to the right place? I had a very frustrating 15 minutes before I realised I wasn't installing where the server was looking. I'm also on Linux, and this does resolve it for me.

> why did you do the handle_control thing?

Can you expand a bit? I'm not quite sure what you're asking.

> could you please add me as owner to the IRkernel team? i’d like to create labelsfor bugs, change urls and descriptions, …

Done.

flying-sheep:
> Are you definitely installing it to the right place? I had a very frustrating 15 minutes before I realised I wasn't installing where the server was looking.

what did you install wrongly? IRkernel itself?

for me, your changes only work once i insert `signal(SIGINT, handle_interrupt)` [before the poll call in C](https://github.com/snoweye/pbdZMQ/blob/29113b7dc6f9fa9c3b9a2bfbff32d4fc20e18af7/src/R_zmq_poll.c#L21) (with `void handle_interrupt(int _) {}`.

> Can you expand a bit? I'm not quite sure what you're asking.

sorry i saw something in the diff wrong. pls ignore.

> Done.

thanks! :smile:

takluyver:
> what did you install wrongly? IRkernel itself?

Yep, I was installing it into the system R install, while I have the kernelspec installed for R in a conda environment (because that's a newer version of R). Maybe the version of R matters? I was testing with 3.2.2.

> for me, your changes only work once i insert signal(SIGINT, handle_interrupt) before the poll call in C (with void handle_interrupt(int _) {}.

I guess this would need to be undone before running user code, though - otherwise interrupting probably won't work when you want it to.

flying-sheep:
still i don’t understand why it would work for you and not me.

i have R 3.2.3…

flying-sheep:
…which is obviously since the python code actually sends the interrupt to R.

i have no idea why your setup actually catches the interrupt while mine doesn’t.

adding a signal handler, calling zmq_poll and removing it works, i’ll take this to pbdZMQ.

---

in any case you should add `suppressWarnings` around `zmq.poll`, else we get

```
Warning message:
In zmq.poll(c(sockets$hb, sockets$shell, sockets$control), rep(.pbd_env$ZMQ.PO$POLLIN,  :
  R_zmq_poll: 4 strerror: Interrupt during System call
```

flying-sheep:
i’ll try the fixes in snoweye/pbdZMQ#25

rolweber:
Ping?

flying-sheep:
Oh, I already did this in 771c6fb5309d6bd0fc162a73058ebc59dda46389

PR #247, opened 2016-01-25
Try installing pbdZMQ on Travis

takluyver:


flying-sheep:
you should also remove the `r-rzmq` from the `conda install` call

takluyver:
Myeh: `error: ISO C++ 1998 does not support ‘long long’ [-Werror=long-long]`

flying-sheep:
hmm, according to jedisct1/libsodium#104, that won’t be changed, so maybe we can do `-Wno-error=long-long`?

takluyver:
OK, it looks like without needing to install rzmq, we can install a newer zmq and libsodium, and the error goes away :-)

flying-sheep:
ok cool, then i’ll do `withCallingHandlers(..., warning = stop)`

Issue #225, opened 2015-12-09
Two [Out] per one [In] cell?


JanSchulz:
As far as I know, in python you can only get one out for each in, but I just managed two outs for one in in R:

```
library(MatchIt) # load matching package
summary(lm(re78~treat,data=lalonde))
t.test(lalonde$re78[lalonde$treat==1],lalonde$re78[lalonde$treat==0],paired=FALSE)
```

![2015-12-09_141953](https://cloud.githubusercontent.com/assets/890156/11686304/fc9cc348-9e7f-11e5-8f99-eebc16e6bbfb.png)

I'm actually not sure what is better, but IMO it should be consistent...

flying-sheep:
jup, we output everything, not just the last expression.

i think that fits R really well since R has `invisible()`, which can be used to

1. return things from functions invisibly, for the exact reason as to prevent it from showing accidentally
2. hide some output you want to hide.

you can just do `invisible(summary(...))` (although i don’t really understand why one would call `summary()` without assigning or displaying the result)

also R is pretty functional in design so this also makes sense. you rarely just call something for calling it, but usually for its return value.

flying-sheep:
so i don’t think it should be done consistently across languages, as language REPLs behave differently and languages have different philosophies.

python | R
----|----
we have side effects | other than writing files, side effects are rare
we often ignore return values | we use or explicitly hide return values
we are OO and therefore mutate state of objects by calling methods | we have no bound methods* and are functional, so we use functions on handles or COW objects

*apart from ref classes

takluyver:
I'm actually a bit surprised by that behaviour - I guess it's a consequence of the way evaluate handles things. It's our intention with the messaging protocol that `execute_result` messages are one per execution, though I guess we'll never strictly enforce that.

If we can send several of them, maybe they should be going as display_data instead of execute_result. Those message types are largely equivalent, but only execute_result results in the `Out[n]` prompt. The idea of those prompts comes from IPython making the result from an execution available for later code as `Out[n]`; since we don't do anything equivalent in IRkernel, it's arguably not important to show them.

flying-sheep:
hmm, semantically, the way things are done now makes the most sense: there are multiple outputs per cell.

that there is only one in IPython seems to be an opinionated choice, as python doesn’t elsewise work like that (i.e. that the last statement in a block/function/code-unit is of special significance).

[...]

PR #251, opened 2016-01-27
Use display_data messages for output instead of execute_result

takluyver:
Closes gh-225

This would result in no `Out[n]:` prompts appearing.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -189,8 +189,7 @@ execute = function(request) {
                 
                 send_response('display_data', request, 'iopub', list(
                     data = data,
-                    metadata = namedlist(),
-                ))
+                    metadata = namedlist() ))
             }
         }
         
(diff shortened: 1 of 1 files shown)


JanSchulz:
I'm +1 on this... But:

```
>>>head(full_df)
Error in list(data = data, metadata = namedlist(), ): argument 3 is empty
Traceback:

1. send_response("display_data", request, "iopub", list(data = data, 
 .     metadata = namedlist(), ))
```

flying-sheep:
yeah, you have to delete the trailing comma.

and i believe our coding style is to have the closing parentheses on the same line

takluyver:
Right, too used to Python.

flying-sheep:
love that in python. [js will have it soon](http://kangax.github.io/compat-table/es7/#test-trailing_commas_in_function_syntax)

Issue #232, opened 2015-12-16
Stop execution of queued cells on error


JanSchulz:
Currently the notebook with an R kernel keeps executing cells even if a cell got an error. In python the currently queued cells are not anymore executed.

eg the following notebook with three cells:

```r
print("whatever")
# --- next cell ---
stop("Stop execution here")
# --- next cell ---
print("Should not run")
```

and then execute the whole notebook via "Cell -> Run All. -> It should only execute the first two cells but not the third. Currently it executes all three.

IMO the current behaviour is very problematic as it can damage data: e.g. when doing a longrunnign computation in cell 2 and saving it to disk in cell 3. Now cell 3 is executed and overwrites the previous data with nothing / wrong things.

JanSchulz:
So what should be happening here? This is currently relaly making me mad, as I use exceptions/stop in my "Kernel>restart & run all" workflow :-(

takluyver:
Meh, it looks like the execution messages all get sent at once, and the kernel is responsible for cancelling them on an exception. That's a bit annoying, but there we go. The Python code that does that is here: https://github.com/ipython/ipykernel/blob/1d97cb2a04149387a0d2dbea1b3d0af691d8df6c/ipykernel/kernelbase.py#L623

We'll need to write an R equivalent.

JanSchulz:
@takluyver This should definitely by one of your kernel test cases :-)

JanSchulz:
PR in https://github.com/IRkernel/IRkernel/pull/264... I will use that tomorrow in some R analysis, lets see if it works...

PR #264, opened 2016-02-22
Abort queued messages if an error occurs

JanSchulz:
The notebook send all execution requests without waiting for the reply and
expects the kernel to abort requests if an error occurs (that's at least the
behavior of the python kernel).

After this commit, the R kernel aborts all shell messages (where execution
requests come in) when the kernel was either interrupted or an error occurred
during the execution of some code.

Fixes: https://github.com/IRkernel/IRkernel/issues/232

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -63,6 +63,7 @@ Executor <- setRefClass(
     'Executor',
     fields = list(
         send_response      = 'function',
+        abort_queued_messages = 'function',
         execution_count    = 'integer',
         payload            = 'list',
         err                = 'list',
@@ -258,7 +259,13 @@ execute = function(request) {
     }
     
     send_response('execute_reply', request, 'shell', reply_content)
-    
+
+    if (interrupted || !is.null(err$ename)){
+        # errors or interrupts should interrupt all currently queued messages,
+        # not only the currently running one...
+        abort_queued_messages()
+    }
+
     if (!silent) {
         execution_count <<- execution_count + 1L
     }
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -91,7 +91,7 @@ new_reply = function(msg_type, parent_msg) {
     
     header <- list(
         msg_id   = UUIDgenerate(),
-        username = parent_msg$header$username, 
+        username = parent_msg$header$username,
         session  = parent_msg$header$session,
         msg_type = msg_type,
         version  = '5.0')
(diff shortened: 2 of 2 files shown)


R/kernel.r
```diff
@@ -129,6 +129,35 @@ handle_shell = function() {
...
+
+abort_queued_messages = function(){
+    "Abort all already queued shell messages after an error"
+
+    while (TRUE) {
+        ret = zmq.poll(
+            c(sockets$shell), # only shell channel
+            c(.pbd_env$ZMQ.PO$POLLIN), # type
+            0) # no timeout, only what's there
+        if (ret != 1){
```
JanSchulz: Not sure if that't the correct way to break here, but for my test case, it was always `1` for the other requests and `0` afterwards.

R/kernel.r
```diff
@@ -129,6 +129,35 @@ handle_shell = function() {
...
+            #  no more messages received
+            break
+        }
+        if(bitwAnd(zmq.poll.get.revents(1), .pbd_env$ZMQ.PO$POLLIN)) {
+            handle_abort()
+        }
```
JanSchulz: ... The above break could probably also go to this if as an `else` clause... Not sure though, I didn't understand teh ZMQ stuff good enough to make this decision :-/

R/execution.r
```diff
@@ -63,6 +63,7 @@ Executor <- setRefClass(
     'Executor',
     fields = list(
         send_response      = 'function',
+        abort_queued_messages = 'function',
```
JanSchulz: R is weird... You have to pass in the function to make them available in a object? Took a while to find this place...

R/kernel.r
```diff
@@ -129,6 +129,35 @@ handle_shell = function() {
         print(c('Got unhandled msg_type:', msg$header$msg_type)))
 },
 
+handle_abort = function(){
```
takluyver: Can we call this something like `abort_shell_msg`? `handle_x` usually means that there is an x arriving that we need to deal with.

R/kernel.r
takluyver: According to the docs, zmq.poll should return two values: 'a ZMQ code and an errno'. So I suspect comparing it to a single integer is not the right thing to do. I'd be inclined to do it as an `else` clause of the bit below, that way there's only one thing we're checking instead of two.

JanSchulz: I read that too, the `1 vs 0` is from `print(ret)`... :-/

R/kernel.r

R/kernel.r
takluyver: The check below is, unless I'm misunderstanding it, the standard 'are there any messages ready on the shell socket' - the shell socket being input number 1 (silly 1-based indexing) to zmq.poll above.

- `zmq.poll.get.revents(1)` -> which events did we get for socket number 1
- `.pbd_env$ZMQ.PO$POLLIN` -> The 'messages to read' event
- `bitwAnd` -> do the events include POLLIN?

JanSchulz: Ok, added the new check...

JanSchulz:
The new check for "empty queue" was actually fixing a bug: it seems that in some cases the first check for the return code wasn't terminating and therefore any new code cell wasn't executed either... At least I never got a result for the next cell after a library failed to load.

takluyver:
Thanks, I think this looks good. And yeah, go ahead and make an issue on jupyter_kernel_test for abort behaviour.

R/execution.r
flying-sheep: reference classes are weird. they’re “OO like it works in java, added as an afterhought”

R/execution.r
```diff
@@ -258,7 +259,13 @@ execute = function(request) {
     }
     
     send_response('execute_reply', request, 'shell', reply_content)
-    
+
+    if (interrupted || !is.null(err$ename)){
```
flying-sheep: space between `) {`

JanSchulz: done

R/execution.r
JanSchulz: I formatted it, but I value proper names more than short lines...

R/execution.r
flying-sheep: thanks!

Issue #265, opened 2016-02-23
Kernel does not take new execution requests after an error


JanSchulz:
Somehow it looks like the error in https://github.com/IRkernel/IRkernel/pull/264 which I thought fixed by the different check for "no more queued messages" is still there:

In a few cases, after an error, I can't submit new execution requests anymore. The notebook tells me it submitted it (execution count is `[*]`), but the result never arrives and every new submission fails as well.

I'm not sure what is triggering it, I can't actually reproduce it :-(

JanSchulz:
Found it in #266: it seems you ar enot allowed to `zmq.poll(3 channels) -> zmq.poll(1 channel) and handle it... -> go back to teh cointext fof the first poll and look if the third channel has something in it`... state machines instead of returned data is fun...

Issue #119, opened 2015-05-01
introduce logging


JanSchulz:
Currently we have some errors which are hard to debug. It also would be nice to tell users a switch and then get some logging output delivered in the console or a file.

from some short searching, I found https://github.com/smbache/loggr

JanSchulz:
I also added a issue to build a debugging lag into the kernel commandline: https://github.com/jupyter/jupyter_client/issues/11

flying-sheep:
also:

* https://github.com/zatonovo/futile.logger
* https://github.com/mfrasca/r-logging

flying-sheep:
what i like in a logger are

1. namespaces (kernel log? executor log?) via logging objects or somehow automatically
2. wrapping/catching native facilities (in R: `warning` und `message`)
3. also if/when we go in CRAN, we’ll want a package on CRAN.

---

loggr und logr seem to do 2., while all other do 1. log4r, futile.logger and logging do 3.

flying-sheep:
so there’s still no package that has both flexible logging objects and integration with `message`, `warning`, and `stop`.

there’s two approaches:

1. `logr` uses the R internal machinery for its stuff and requires the [`with_logging`](https://github.com/cubranic/logr/blob/b003a45c16e08fb9e4e9d1fbb223df570fdbd20c/R/logging.R#L52-L76) wrapper to reroute those to a file.
2. `loggr` pokes around in the interna and [makes *them* use *its* machinery](https://github.com/smbache/loggr/blob/master/R/set_loggr_hook.R).

and still, both don’t use multilevel logging/logging objects.

janschulz:
> wrapping/catching native facilities (in R: warning und message)

Don't we anyway have to do our own catching? At least in the evaluate call and probably also in the kernel itself, so that we don't break on bad inputs...

flying-sheep:
nonono, not `stop` catching. i meant that code which relies on `message` and `warning` interacts with the logging system and not stupidly prints to stderr.

janschulz:
> nonono, not stop catching. i meant that code which relies on message and warning interacts with the logging system and not stupidly prints to stderr

We would still need to catch these stuff in the user supplied code to get this to the frontend... So this helps us only with our own code in the kernel/repr/irdisplay and maybe evaluate, which AFAIK all do not use message/warning?

flying-sheep:
display and repr use nothing (both pretty slim and testable, no infinite loops, …)

[...]

PR #266, opened 2016-02-24
Fix hang after an aborted message

JanSchulz:
This fixes two bugs introduced by 61f0ca492e7a98c57d3c31b2020ad16642fe00e4:

* The kernel would hang in some circumstances:  It seems that zmq can't handle
  the case where the results of the first poll haven't been handled and and
  another poll is already started and the handling of the first poll is then
  resumed. In this case the main loop thought it found a control message and
  started handling it, but the function never returned from the
  `zmq.recv.multipart`. The workaround here was reordering the handle_xxx flow
  to let shell messages be handled last.

* The reply message had an incorrect type, because I forgot an unlist()

This also introduces a debug print function which does nothing and has to be activated to print to stderr...

Closes: https://github.com/IRkernel/IRkernel/issues/265

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -7,6 +7,10 @@ fromRawJSON <- function(r) {
     fromJSON(s)
 }
 
+# Uncomment to get debug printouts to stderr
+#debug <- function(...) cat(sprintf(...), '\n', sep = '', file = stderr())
+debug <- function(...){}
+
 #' The kernel
 #' 
 #' Has methods able to connect and talk to a Jupyter server.
@@ -111,6 +115,7 @@ send_response = function(msg_type, parent_msg, socket_name, content) {
     msg$content <- content
     socket <- sockets[[socket_name]]
     zmq.send.multipart(socket, msg_to_wire(msg), serialize = FALSE)
+    debug('Sending msg %s', msg$header$msg_type)
 },
 
 handle_shell = function() {
@@ -129,33 +134,41 @@ handle_shell = function() {
         print(c('Got unhandled msg_type:', msg$header$msg_type)))
 },
 
-abort_shell_msg = function(){
+abort_shell_msg = function() {
     "Send an abort message for an incoming shell request"
     # See https://github.com/ipython/ipykernel/blob/1d97cb2a04149387a0d2dbea1b3d0af691d8df6c/ipykernel/kernelbase.py#L623
 
     parts <- zmq.recv.multipart(sockets$shell, unserialize = FALSE)
     msg <- wire_to_msg(parts)
-    reply_type <- paste(strsplit(msg$header$msg_type, "_")[1], "_reply")
+    debug('Aborting msg of type %s', msg$header$msg_type)
+    reply_type <- paste0(unlist(strsplit(msg$header$msg_type, '_'))[1], '_reply')
     reply_content <- list(status = 'aborted')
-    send_response(reply_type, msg, "shell", reply_content)
+    send_response(reply_type, msg, 'shell', reply_content)
+    debug('Aborted msg')
 },
 
-abort_queued_messages = function(){
+abort_queued_messages = function() {
     "Abort all already queued shell messages after an error"
 
+    debug('abort loop: aborted all outstanding msg')
     while (TRUE) {
+        debug('abort loop: before poll')
         ret = zmq.poll(
             c(sockets$shell), # only shell channel
             c(.pbd_env$ZMQ.PO$POLLIN), # type
             0) # zero timeout, only what's already there
-
+        debug('abort loop: after poll')
         if(bitwAnd(zmq.poll.get.revents(1), .pbd_env$ZMQ.PO$POLLIN)) {
+            debug('abort loop: found msg')
             abort_shell_msg()
         } else {
             # no more messages...
+            debug('abort loop: breaking')
             break
         }
     }
+    debug('abort loop: end')
+
 },
 
 is_complete = function(request) {
@@ -256,9 +269,12 @@ kernel_info = function(request) {
 },
 
 handle_control = function() {
+    debug('Control: beginning')
     parts <- zmq.recv.multipart(sockets$control, unserialize = FALSE)
     msg <- wire_to_msg(parts)
+    debug('Control: recv msg of type %s', msg$header$msg_type)
     if (msg$header$msg_type == 'shutdown_request') {
+        debug('Control: shutdown...')
         shutdown(msg)
     } else {
         print(paste('Unhandled control message, msg_type:', msg$header$msg_type))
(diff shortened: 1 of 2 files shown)


JanSchulz:
IMO its a bug that ` bitwAnd(zmq.poll.get.revents(3), .pbd_env$ZMQ.PO$POLLIN)` can be true after a `zmq.poll(...)` with only a single channel :-(

R/kernel.r
```diff
@@ -135,27 +139,34 @@ abort_shell_msg = function(){
 
     parts <- zmq.recv.multipart(sockets$shell, unserialize = FALSE)
     msg <- wire_to_msg(parts)
-    reply_type <- paste(strsplit(msg$header$msg_type, "_")[1], "_reply")
+    debug("Aborting msg of type %s", msg$header$msg_type)
+    reply_type <- paste(unlist(strsplit(msg$header$msg_type, "_"))[1], "_reply", sep="")
```
flying-sheep: `paste0(...)` is the better way to do `paste(..., sep = '')`

R/kernel.r
```diff
@@ -7,6 +7,9 @@ fromRawJSON <- function(r) {
     fromJSON(s)
 }
 
+#debug <- function(...) cat(sprintf(...), "\n", sep='', file=stderr())
+debug <- function(...){}
```
flying-sheep: we should fix #119 instead

R/kernel.r
```diff
@@ -111,6 +114,7 @@ send_response = function(msg_type, parent_msg, socket_name, content) {
     msg$content <- content
     socket <- sockets[[socket_name]]
     zmq.send.multipart(socket, msg_to_wire(msg), serialize = FALSE)
+    debug("Sending msg %s", msg$header$msg_type)
```
flying-sheep: this and everywhere else: we use `'`, not `"`

flying-sheep:
thanks for this!

we should file an issue about `zmq.poll.get.revents(3)` in pbdZMQ

R/kernel.r
takluyver: So long as it's easy to point the debug function to proper logging later, I don't mind this as an interim solution. The perfect is the enemy of the good and all that.

R/kernel.r
takluyver: I continue not to care, but I'll defer to you on style questions.

R/kernel.r
```diff
@@ -300,20 +314,30 @@ initialize = function(connection_file) {
 
 run = function() {
     while (TRUE) {
+        debug("main loop: beginning")
         zmq.poll(
             c(sockets$hb, sockets$shell, sockets$control),
             rep(.pbd_env$ZMQ.PO$POLLIN, 3))
-        
-        if(bitwAnd(zmq.poll.get.revents(1), .pbd_env$ZMQ.PO$POLLIN))
```
takluyver: Rather than just reordering them, which could cause a similar issue if we ever call poll on a control message, I think we should use `next` ([equivalent](http://code.hammerpig.com/loops-continue.html) to `continue` in Python) to jump back to the polling call after handling any message. That way, we never try to resume handling a poll result after dispatching messages.

takluyver: Though we'll need to keep the reordering in that case so that control messages can take priority over shell messages.

R/kernel.r
JanSchulz: Will fix in the evening...

R/kernel.r
JanSchulz: As long as a new poll does not remove an already stored message, a `if ... Else if ... Else ...` should also work... Will try later...

R/kernel.r
flying-sheep: OK

R/kernel.r
flying-sheep: thanks!

R/kernel.r

R/kernel.r
JanSchulz: I used a big `if ... else if ... else if ... else <should not happen>`, which seems to work...

JanSchulz:
Doe alle the changes (I hope)...

Anyone against remoing all trailing whitespace? There are a few empty lines which have spaces and my editor removes this on save and then I have to use `git add -p` to not introduce them into the diff...

takluyver:
Thanks!

I am generally against trimming all the trailing whitespace, because it's unnecessary diff noise - it can create spurious conflicts and make git blame less useful. We ask Jupyter/IPython contributors to redo commits where their editor has automatically stripped whitespace. Most editors can be configured not to, and some can even trim only on lines that you've changed.

However, Philipp calls the shots on code style for IRkernel, so it's up to him what we do with whitespace.

JanSchulz:
@takluyver Thats why I didn't add the strips :-) But usually all whitespace is stripped, at least in the jupyter repos :-)

flying-sheep:
usually i configure my editor to only strip (or add) automatic stylistic changes on lines i edited anyway.

and i don’t see any trailing whitespace anyway. do you mean indented empty lines? why should they be unindented when the surrounding lines are indented?

flying-sheep:
about the actual content: this is tested by simply creating a few cells, one of which calls `stop()`, and subsequent cells should not execute right? if so, i’ll test and report back.

we could make the `debug` stuff more useful via `debug <- if (Sys.getenv('DEBUG') != '') function(...) message(sprintf(...)) else function(...) {}`

takluyver:
> about the actual content: this is tested by simply creating a few cells, one of which calls stop(), and subsequent cells should not execute right? if so, i’ll test and report back.

Yep, if they're already queued - for instance, if you've done 'run all'. After the error, it should still be possible to queue later cells and have them execute.

Let's leave the debugging stuff as it is for this PR, and work that out in a separate PR. It's orthogonal to the change being discussed here.

JanSchulz:
> usually i configure my editor to only strip (or add) automatic stylistic changes on lines i edited anyway.

Didn't know that was possible, but as always: there is a plugin for that :-)

> and i don’t see any trailing whitespace anyway. do you mean indented empty lines? why should they be unindented when the surrounding lines are indented?

There were a few trialing spaces after a normal line, but most of them were empty lines like the line after a function docstring:

```
fun <- function(){
    "text"
    
    print("x")
```

Here the line after "text" was filled with 4 spaces... In python I would remove this (more or less because I was forced to by the pep8 tests in matplotlib which made my install a trailing ws remover...)

JanSchulz:
I've now used this version of the code during the last few hours and didn't get a deadlocked kernel, so it seems this is working now :-)

takluyver:
Thanks :-)

flying-sheep:
> There were a few trialing spaces after a normal line, but most of them were empty lines like the line after a function docstring

those are only technically “trailing”, but in fact belong to the code block indentation:

code|semantics
----|----
`fun <- function(){` | 0 indentation, 0 trailing whitespace
`    "text"` | 4 spaces indentation, 0 trailing whitespace
`    ` | 4 spaces indentation, 0 trailing whitespace
`      ` | 4 spaces indentation, 2 trailing spaces
`    print("x") ` | 4 spaces indentation, 1 trailing space

JanSchulz:
@flying-sheep Although I disagree about some specifics (and that you always find them and I have to fix it :-) ), I do find a uniformly formatted codebase a good thing!

Anyway: found this today and found it appropriate :-)

http://devopsreactions.tumblr.com/post/140324458371/when-someone-doesnt-follow-the-coding-style

flying-sheep:
haha i’m that cat right?

sorry!

if you want, we can do it this way in the future:

1. delete your fork
2. create feature branches in this repo
3. i’ll fix stuff that peeves me directly in the branch while the PR is open

JanSchulz:
:-)

No, I'm fine with fixing whatever mistakes I make... Is there a automatic code checker for R (like pep8 for python)?

flying-sheep:
yes: [lintr](https://github.com/jimhester/lintr)

it’s good, but unfortunately has many false positives since it doesn’t know refclasses.

but it catches all the other stuff!

Issue #24, opened 2014-05-14
Set some option so code can detect when it's running in IRkernel


takluyver:
From a discussion with @ramnathv.

Note that being in IRkernel does not necessarily imply that the code is running in the notebook - there are multiple frontends speaking the same protocol, and they may even be connected to the kernel at the same time. But we can reliably indicate that we're in the kernel.

flying-sheep:
@JanSchulz said:

> My proposal:
> 
> in the kernel.r$run add an option:
> 
> ```r
> option(irkernel.in_kernel = TRUE)
> ```
> 
> Packages can then check for this and adjust behaviour (e.g. produce different markdown output)
> 
> ```r
> getOption("irkernel.in_kernel", FALSE)
> ```
> 
> Any objections? Different names?

and we should set it to `FALSE` in the options file so users can discover it.

flying-sheep:
1. `jupyter.in_kernel` to match `jupyter.plot_mimetypes`
2. [set it to `FALSE` here](https://github.com/IRkernel/IRkernel/blob/ba45b39767be483a3da2bea0eb77e8fac2243aa1/R/options.r#L19)

flying-sheep:
1. `jupyter.in_kernel` to match `jupyter.plot_mimetypes`
2. [set it to `FALSE` here](https://github.com/IRkernel/IRkernel/blob/ba45b39767be483a3da2bea0eb77e8fac2243aa1/R/options.r#L19)

JanSchulz:
IMO this isn't worth to set to false: who will import the r kernel in a regular session? And users of this will anyway have to handle the case that it isn't set because in the majority of the cases (=RStdio users...) it won't.

flying-sheep:
it’ll be more visible that way.

PR #272, opened 2016-03-01
Added Comm Message Support

poplav:
This is our first stab at adding Comm support to the R Kernel. Please give us feedback and we'll make any changes necessary.

We are actually using this fork to implement the support for declarativewidgets under https://github.com/jupyter-incubator/declarativewidgets/pull/244.


Comm_Demo.ipynb
```diff
@@ -0,0 +1,185 @@
+{
```
flying-sheep: when we get so many example notebooks, maybe it would be best to move all 4 into a common directory.

be sure to add the directory to .RBuildignore

R/comm_manager.r
```diff
@@ -0,0 +1,186 @@
...
+    'Comm_Manager',
+    fields = list(
+        send_response = 'function',
+        target_to_handler_map = 'list',
+        commid_to_comm = 'list',
+        parent_request = 'list'
+    ),
+    methods = list(
+        new_comm = function(target_name, comm_id=NULL){
+            if( is.null(comm_id) ){
```
flying-sheep: and this should read `if (is.null(comm_id)) {`

R/comm_manager.r
```diff
@@ -0,0 +1,186 @@
...
+                comm_id = UUIDgenerate()
+            }
+            return (Comm$new(id=comm_id, target_name=target_name, comm_manager=.self))
```
flying-sheep: no parentheses, please

lbustelo:
@flying-sheep Appreciate the feedback... you have here a couple of Java and Javascript devs venturing into the land of R. Keep it coming!

R/kernel.r
```diff
@@ -312,6 +316,8 @@ initialize = function(connection_file) {
 
     executor <<- Executor$new(send_response = .self$send_response,
         abort_queued_messages = .self$abort_queued_messages)
+    comm_manager <<- Comm_Manager$new(send_response = .self$send_response)
+    assign('kernel', .self, envir = .GlobalEnv)
```
lbustelo: No... but we welcome other ideas. We needed to have access to the kernel and comm_manager within the interpreter. Or first stab at it was to add the kernel to the environment.

lbustelo: See the attached notebook for examples of how to use the API from a code cell.

jdfreder:
cc @sylvaincorlay

lbustelo:
Missing support for http://jupyter-client.readthedocs.org/en/latest/messaging.html#comm-info

R/comm_manager.r
```diff
@@ -0,0 +1,186 @@
...
+                # create a comm object
+                comm <- new_comm(target_name, comm_id)
+                register_comm(comm)
+                data <- request$content$data
+
+                #invoke target handler
+                tryCatch({
+                    target_to_handler_map[[target_name]](comm, data)
+                }, error = function(e) {
+                    print("error invoking the handler for target")
```
JanSchulz: https://github.com/IRkernel/IRkernel/blob/master/R/kernel.r#L10-L12, but this is currently more than hacky... :-/

jdfreder:
@poplav looks like you haven't added the email in 2226d96 to your account, so you won't get credit for it.  Did you want to change that?

R/comm_manager.r
flying-sheep: it’s OK, we’ll replace it once #119 is fixed (damn, i know that number by heart right now…)

R/kernel.r
flying-sheep: this is related to #24.

if we do it this way, a valid check would be `inherits(.GlobalEnv$kernel, 'kernel')`.

another possibility would be to have a global kernel instance in the `IRkernel` namespace, which is usually `NULL` unless we run in a kernel.

maybe better than modifying the global env, what do y’all think?

Comm_Demo.ipynb
```diff
@@ -0,0 +1,185 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Test in Javascript console\n",
+    "var comm = IPython.notebook.kernel.comm_manager.new_comm(\"kernel_target\",{list:[1,2,3,4,5,6,7,8,9]})\n",
+    "\n",
+    "comm.send({foo:'bar'})\n",
+    "\n",
+    "comm.close()"
```
flying-sheep: just do this:

```R
display_javascript(sprintf('
var comm = IPython.notebook.kernel.comm_manager.new_comm("%s",{list:[1,2,3,4,5,6,7,8,9]})

comm.send({foo:"bar"})

comm.close()
', target_name_in_kernel))
```

R/kernel.r
flying-sheep: also if the only use of the kernel object is to provide an interface to adding and removing comm managers, we should instead expose an object that allows us to do only that.

i suggest

1. documenting `CommManager` and `Comm` by `#'@export` before the class and adding docstrings to methods
2. only making the `comm_manager` available, not the kernel

Comm_Demo.ipynb
poplav: Nice, thanks.

poplav:
119c7f6, should resolve issues by:
* Exposing comm_manager not kernel to the global environment
* Adding support for [Comm Info requests](http://jupyter-client.readthedocs.org/en/latest/messaging.html#comm-info)
* Moving our Comm_Demo notebook to the example notebook section
* debug statements/various styling issues

Also where are we and the IRkernel project as whole at with regard to tests?  Adding a test for say comm_info_request would be nice?  Any pointers?

flying-sheep:
thank you!

the only testing we do yet is [`test_ir.py`](https://github.com/IRkernel/IRkernel/blob/master/test_ir.py).

i think it doesn’t allow testing comms, but maybe you can do things via [`code_execute_result` and `code_display_data`](https://github.com/jupyter/jupyter_kernel_test)

else we need to come up with something that boots up a kernel in the background and sends arbitrary stuff to it.

lbustelo:
For those types of integration tests, I wonder if https://github.com/jupyter/jupyter_kernel_test would help. 

How about for unit testing?

flying-sheep:
`test_ir.py` uses `jupyter_kernel_test`.

we’d love to have unit tests! first we need code that spins up and easily communicates with a server…

takluyver:
jupyter_kernel_test does spin up a kernel! You can attach any further test methods you want to the subclass, e.g. `test_irkernel_comms()` (I'd suggest putting `irkernel` or something distinctive in the name, so that it won't clash with any tests JKT might add later).

Inside the method, `self.km` and `self.kc` are a kernel manager and kernel client, respectively. There's also an [execute_helper()](https://github.com/jupyter/jupyter_kernel_test/blob/b51d9911bc7a8bfbb826d3fd3149767c319b73fe/jupyter_kernel_test/__init__.py#L59) method which sends some code to be executed and collects all the output messages the kernel produces.

takluyver:
(This is a major reason why you use JKT by subclassing it in Python code - so you can easily define extra tests for this kind of situation.)

flying-sheep:
could you please add an example test which @poplav or others can use as a template for their tests?

poplav:
When we first opened this PR we said we were using it to implement R support in [jupyter-incubator/declarativewidgets](https://github.com/jupyter-incubator/declarativewidgets).  We have made significant progress under https://github.com/jupyter-incubator/declarativewidgets/pull/244 and would like to share a demo:
urth-r-widgets-demo
What else is needed to merge this PR?  Thanks!

flying-sheep:
nothing, this looks great! just details like whitespace-only changes, and the docs being sparse, but we can do that later

@takluyver your opinion?

flying-sheep:
oh, and we should add a function:

```r
in_irkernel <- function() {
    .GlobalEnv$comm_manager != NULL
}
```

JanSchulz:
Knitr uses an option for that, but I guess and option can't be set based on this object :-/

flying-sheep:
well, we can also set an option of course, which would give people the opportunity to fake being in a kernel. but i don’t see why that would be necessary.

let’s stick with the function. we can always change it to take an option into account.

takluyver:
I looked over this before and didn't see anything obviously amiss. I'm off to Amsterdam for a conference over the weekend, so don't hold this up waiting for me

JanSchulz:
An option has the nice thing, that you don't need to catch the errors when the function is not available...

E.g. if pander (see https://github.com/IRkernel/repr/issues/33#issuecomment-194259812) should set the default output to `rmarkdown` if it is in the kernel.

flying-sheep:
anything using the function IRkernel with the function present should simply add `Imports: IRkernel>=0.7` or so. that way people will see they have to upgrade to a version having the function.

JanSchulz:
> anything using the function IRkernel with the function present should simply add Imports: IRkernel>=0.7

I'm not sure if there is any magic behind this, but if this downloads irkernel when someone installs `pander` just to let pander see that we are not in the kernel, then I would be pissed as a user for that waste of space.

```
getOption("irkernel.in_a_kernel", FALSE)
```

flying-sheep:
good point.

poplav:
@flying-sheep: 28c213b, exposes a comm_manager function which allows us to access the actual comm_manager via the comm_manager_env that gets created.  We then get access to the comm in our example notebook via:  ```library(IRkernel)
    comm_manager <- get("comm_manager", envir = as.environment("package:IRkernel"))()```

flying-sheep:
hey, many thanks! it all works great!

but you can just call `comm_mgr <- comm_manager()`, no need for the `get` call!

poplav:
:tada:

flying-sheep:
very cool stuff!

SylvainCorlay:
:+1: :tad: :fireworks:

SylvainCorlay:
If we were to write a R backend for the core Jupyter widget models, what would be a good counterpart to traitlets?

poplav:
@SylvainCorlay, this example may be insightful http://r.789695.n4.nabble.com/monitor-variable-change-tt3308460.html#a3308935

PR #276, opened 2016-03-09
Add flag to show when we are running in a kernel

JanSchulz:
We now set an options flag when we are running a kernel. This can be used to
trigger behavior changes in other packages, like using a different markdown
format when outputting markdown.

Example:

```
if (getOption("jupyter.in_kernel", FALSE)){
	# Code which should only be running when in a kernel
}
```

Closes: https://github.com/IRkernel/IRkernel/issues/24


flying-sheep:
:smile: thank you!

PR #284, opened 2016-03-23
Handle errors in display functions more gracefully

janschulz:
Before, when one repr_xxx was raising an error, the complete display machinery
was impacted and no output was displayed. Now only the specific output is
omitted and a user friendly error message + stacktrace is shown on stderr and 
on the console.

Closes: https://github.com/IRkernel/IRdisplay/issues/18

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -182,15 +182,19 @@ execute = function(request) {
         # right now.
         handle_display_error <- function(e){
             # This is used with withCallingHandler and only has two additional
-            # calls at the end instead of the 3 for tryCatch...
-            calls <- head(sys.calls()[-seq_len(nframe + 1L)], -2)
+            # calls at the end instead of the 3 for tryCatch... (-2 at the end)
+            # we also remove the tryCatch and mime2repr stuff at the head of the callstack (+7)
+            calls <- head(sys.calls()[-seq_len(nframe + 7L)], -2)
             stack_info <- format_stack(calls)
             # TODO: replace with proper logging
-            cat(sprintf('ERROR: %s\nTraceback:\n%s\n', toString(e), paste(stack_info, collapse='\n')), file = stderror)
+            msg <- sprintf('ERROR while rich displaying an object: %s\nTraceback:\n%s\n',
+                           toString(e),
+                           paste(stack_info, collapse='\n'))
+            cat(msg, file = stderror)
             if (!silent) {
                 send_response('stream', request, 'iopub', list(
                     name = 'stderr',
-                    text = sprintf('Error while rich displaying an object: %s', toString(e))))
+                    text = msg))
             }
         }
         handle_value <- function(obj) {
--- a/R/execution.r
+++ b/R/execution.r
@@ -148,6 +148,10 @@ execute = function(request) {
     
     err <<- list()
     nframe <- NULL  # find out stack depth in notebook cell
+
+    # handle to stderr if stderr() is sink'ed
+    stderror <- stderr()
+
     tryCatch(evaluate(
         'stop()',
         stop_on_error = 1L,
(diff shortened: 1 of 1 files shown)


janschulz:
```R
funct <- function(x){
    stop("Not my business")
}
xprint.test_object <- function(x){
    funct(x)
}

repr_html.test_object <- function(x){
    funct(x)
}
t <- structure(paste("test", collapse="\n"), class = 'test_object')
t
```

displays:

```
Error while rich displaying an object: Error in funct(x): Not my business
[1] "test" attr(,"class") [1] "test_object"
```

and in the console is now this:

```
> IRkernel::main()
ERROR: Error in funct(x): Not my business

Traceback:
1. tryCatch(withCallingHandlers({
 .     r <- mime2repr[[mime]](obj)
 . }, error = handle_display_error), error = function(x) {
 . })
2. tryCatchList(expr, classes, parentenv, handlers)
3. tryCatchOne(expr, names, parentenv, handlers[[1L]])
4. doTryCatch(return(expr), name, parentenv, handler)
5. withCallingHandlers({
 .     r <- mime2repr[[mime]](obj)
 . }, error = handle_display_error)
6. mime2repr[[mime]](obj)
7. repr_html.test_object(obj)
8. funct(x)   # at line 9 of file <text>
9. stop("Not my business")   # at line 2 of file <text>
```

R/execution.r
```diff
@@ -173,6 +177,20 @@ execute = function(request) {
...
         handle_warning  <- identity
     } else {
+        # We only want a short error in the notebook/... because the usual case
+        # is just a problem in one of the display methods which are not relevant
+        # right now.
+        handle_display_error <- function(e){
+            calls <- head(sys.calls()[-seq_len(nframe + 1L)], -3)
+            stack_info <- format_stack(calls)
+            # TODO: replace with proper logging
+            cat(sprintf('ERROR: %s\nTraceback:\n%s\n', toString(e), paste(stack_info, collapse='\n')), file = stderror)
```
takluyver: I think the traceback belongs in the output we send to the frontend.

R/execution.r
```diff
@@ -180,12 +198,11 @@ execute = function(request) {
...
-                    tryCatch({
-                        for (mime in getOption('jupyter.display_mimetypes')) {
-                            r <- mime2repr[[mime]](obj)
-                            if (!is.null(r)) data[[mime]] <- r
-                        }
-                    }, error = handle_error)
+                    for (mime in getOption('jupyter.display_mimetypes')) {
+                        tryCatch(r <- mime2repr[[mime]](obj),
+                            error = handle_display_error)
+                        if (!is.null(r)) data[[mime]] <- r
```
takluyver: If the repr fails, I think `r` will either be undefined, or have the value from the previous iteration of the loop. So I think this needs to be inside the tryCatch, so it only runs if the repr succeeds.

takluyver:
GH is deciding that the two comments I just made no longer need to be visible, but they both still stand with the latest change ;-)

janschulz:
> I think the traceback belongs in the output we send to the frontend.

I'm very much against this: that would mean that an error in repr_latex (which essentially will not be shown ever in a normal case -> notebook, console) will always be visible and take up quite a lot of space. Now we have a small oneline error message which could still be ignored... 

A compromise could be to add a "(stacktrace in the kernel log)" to the error message. Would that satisfy you?

R/execution.r
janschulz: true, will change

takluyver:
If it's easily ignored, then errors in repr_latex won't get fixed.

It's also what IPython does, and IPython tracebacks are larger than R tracebacks:
https://gist.github.com/takluyver/a854d51ce047b8ba17b8

janschulz:
ok, fix is underway...

janschulz:
New output:

```
ERROR while rich displaying an object: Error in funct(x): Not my business

Traceback:
1. repr_html.test_object(obj)
2. funct(x)   # at line 9 of file <text>
3. stop("Not my business")   # at line 2 of file <text>
```

takluyver:
Looks OK to me now. I'm not quite sure why the `withCallingHandlers` is needed, but I'll take your word that it is. I'll let @flying-sheep have a look too.

janschulz:
```R
x <- function() { y(); }
y <- function() { z(); }
z <- function() { stop("asdf") }
# ---
print("Does not show the reason for the error, only the outer most function")
tryCatch(x(), error=function(e) { print(sys.calls())} )
# ---
print("Does not catch the error -> kernel shows the stacktrace a second time")
withCallingHandlers(x(), error=function(e) print(sys.calls()))
# and in this case, another statement would not be executed
# ---
print("And the right thing: show the right stacktrace and catches the error")
tryCatch(withCallingHandlers(x(), error=function(e) print(sys.calls())), error=function(x){})
```

janschulz:
All commits are in #285, so close this PR

Issue #127, opened 2015-05-08
Setting data.table column outputs whole table to screen


abielr:
If I run the code below, the second line will cause the entire `dat` object to be output, whereas at an R console it wouldn't return anything. The syntax used is the special syntax for setting columns with the popular `data.table` package. I'm using the 1.9.5 devel version of `data.table`. 

```
library(data.table)
dat <- data.table(x1=1:10)
dat[, x2 := 1:10]
```

takluyver:
`withVisible(d[, x2:=1:10])` tells me that the result should be visible. I thought that was how R determined whether or not to print it. Clearly there's something else going on.

abielr:
Just a note if testing this: the same behavior you see in the notebook is also cropping up at the console in R 3.2.0, as described at [https://github.com/Rdatatable/data.table/issues/1122](). So if trying to fix you will likely want to use R < 3.2.0 until they get this bug fixed.

abielr:
Fixing this on the IRkernel side is probably not reasonable. Under the hood, when you run a statement with the data.table `:=` syntax, it sets a global variable equal to `address(x)`, which is then checked inside of the `print.data.table` command to see if the output should be suppressed. In other words, when you run the command

`dat[, x2 := 1:10]`

It is triggering `[.data.table` followed by `print.data.table`, where `print.data.table` sees that it should not print this time, after which it resets the state of the global variable so that other print statements will run. However, when IRkernel is running `evaluate()`, the `print.data.table` statement never gets called, so all you see is the regular `data.table` object.

takluyver:
Ah, R ;-)

abielr:
Unfortunately because `[` is a R primitive function, it cannot return an invisible object, see [here](http://stackoverflow.com/questions/30246290/make-return-from-s3-indexing-function-invisible). This is why the developers of `data.table` have resorted to the workaround of using an internal global variable to alter the behavior of `print.data.table`.

A fix on the IRkernel side is to check of the length of `data[['text/plain']]` inside the `execution.R/handle_value()` function. If the length is zero, `nchar(data[['text/plain']]==0`, then don't send back a response because if you were running this at the console then you would be getting nothing, and in general we would expect the printing of text objects to mimic the console in the notebook. This still allows the notebook to work properly with functions that return a blank string or NULL, which will print properly. 

[...]

PR #285, opened 2016-03-24
Handle errors in display functions more gracefully and isolate full html pages

janschulz:
Before, when one repr_xxx was raising an error, the complete display machinery
was impacted and no output was displayed. Now only the specific output is
omitted and a user friendly error message + stacktrace is shown on stderr and 
on the console.

Also indicate that full html pages should be isolated. Full html pages should be
isolated in the frontend in an iframe. Check if the html is a complete document
(contains html) and set the right metadata.

Closes: https://github.com/IRkernel/IRdisplay/issues/18
Closes: #284

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -178,7 +178,9 @@ execute = function(request) {
         # is just a problem in one of the display methods which are not relevant
         # right now.
         handle_display_error <- function(e){
-            calls <- head(sys.calls()[-seq_len(nframe + 1L)], -3)
+            # This is used with withCallingHandler and only has two additional
+            # calls at the end instead of the 3 for tryCatch...
+            calls <- head(sys.calls()[-seq_len(nframe + 1L)], -2)
             stack_info <- format_stack(calls)
             log_debug('ERROR: %s\nTraceback:\n%s\n', toString(e), paste(stack_info, collapse='\n'))
             if (!silent) {
@@ -195,10 +197,15 @@ execute = function(request) {
             if (nchar(data[['text/plain']]) > 0) {
                 if (getOption('jupyter.rich_display')) {
                     for (mime in getOption('jupyter.display_mimetypes')) {
-                        tryCatch({
-                            r <- mime2repr[[mime]](obj)
-                            if (!is.null(r)) data[[mime]] <- r
-                            }, error = handle_display_error)
+                        # Use withCallingHandlers as that shows the inner stacktrace:
+                        # https://stackoverflow.com/questions/15282471/get-stack-trace-on-trycatched-error-in-r
+                        # the tryCatch is  still needed to prevent the error from showing
+                        # up outside further up the stack :-/
+                        tryCatch(withCallingHandlers({
+                                r <- mime2repr[[mime]](obj)
+                                if (!is.null(r)) data[[mime]] <- r
+                            }, error = handle_display_error),
+                            error = function(x) {})
                     }
                 }
                 
--- a/R/execution.r
+++ b/R/execution.r
@@ -192,6 +192,8 @@ execute = function(request) {
         }
         handle_value <- function(obj) {
             data <- namedlist()
+            metadata <- namedlist()
+
             data[['text/plain']] <- repr_text(obj)
             
             # Only send a response when there is regular console output
@@ -203,17 +205,26 @@ execute = function(request) {
                         # the tryCatch is  still needed to prevent the error from showing
                         # up outside further up the stack :-/
                         tryCatch(withCallingHandlers({
-                                r <- mime2repr[[mime]](obj)
-                                if (!is.null(r)) data[[mime]] <- r
-                            }, error = handle_display_error),
-                            error = function(x) {})
+                            r <- mime2repr[[mime]](obj)
+                            if (!is.null(r)) {
+                                data[[mime]] <- r
+                                # Isolating full html pages (putting them in an iframe)
+                                if (identical(mime, 'text/html')) {
+                                    if (grepl("<html.*>", r, ignore.case = TRUE)){
+                                        jupyter_debug("Found full html page: %s", strtrim(r, 100))
+                                        metadata[[mime]] <- list(isolated = TRUE)
+                                    }
+                                }
+                            }
+                        }, error = handle_display_error),
+                        error = function(x) {})
                     }
                 }
-                
-                send_response('display_data', request, 'iopub', list(
-                    data = data,
-                    metadata = namedlist() ))
             }
+            log_debug("Sending display_data: %s", paste(capture.output(str(data)), collapse = "\n"))
+            send_response('display_data', request, 'iopub', list(
+                data = data,
+                metadata = metadata))
         }
         
         stream <- function(output, streamname) {
--- a/R/execution.r
+++ b/R/execution.r
@@ -174,19 +174,20 @@ execute = function(request) {
         handle_message  <- identity
         handle_warning  <- identity
     } else {
-        # We only want a short error in the notebook/... because the usual case
-        # is just a problem in one of the display methods which are not relevant
-        # right now.
         handle_display_error <- function(e){
             # This is used with withCallingHandler and only has two additional
-            # calls at the end instead of the 3 for tryCatch...
-            calls <- head(sys.calls()[-seq_len(nframe + 1L)], -2)
+            # calls at the end instead of the 3 for tryCatch... (-2 at the end)
+            # we also remove the tryCatch and mime2repr stuff at the head of the callstack (+7)
+            calls <- head(sys.calls()[-seq_len(nframe + 7L)], -2)
             stack_info <- format_stack(calls)
-            log_debug('ERROR: %s\nTraceback:\n%s\n', toString(e), paste(stack_info, collapse='\n'))
+            msg <- sprintf('ERROR while rich displaying an object: %s\nTraceback:\n%s\n',
+                           toString(e),
+                           paste(stack_info, collapse='\n'))
+            log_debug(msg)
             if (!silent) {
                 send_response('stream', request, 'iopub', list(
                     name = 'stderr',
-                    text = sprintf('Error while rich displaying an object: %s', toString(e))))
+                    text = msg))
             }
         }
         handle_value <- function(obj) {
--- a/R/execution.r
+++ b/R/execution.r
@@ -148,7 +148,7 @@ execute = function(request) {
     
     err <<- list()
     nframe <- NULL  # find out stack depth in notebook cell
-    
+
     tryCatch(evaluate(
         'stop()',
         stop_on_error = 1L,
@@ -174,6 +174,19 @@ execute = function(request) {
         handle_message  <- identity
         handle_warning  <- identity
     } else {
+        # We only want a short error in the notebook/... because the usual case
+        # is just a problem in one of the display methods which are not relevant
+        # right now.
+        handle_display_error <- function(e){
+            calls <- head(sys.calls()[-seq_len(nframe + 1L)], -3)
+            stack_info <- format_stack(calls)
+            log_debug('ERROR: %s\nTraceback:\n%s\n', toString(e), paste(stack_info, collapse='\n'))
+            if (!silent) {
+                send_response('stream', request, 'iopub', list(
+                    name = 'stderr',
+                    text = sprintf('Error while rich displaying an object: %s', toString(e))))
+            }
+        }
         handle_value <- function(obj) {
             data <- namedlist()
             data[['text/plain']] <- repr_text(obj)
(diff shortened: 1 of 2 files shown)


janschulz:
So, the main reason is that this now works:

```R
#devtools::install_github("renkun-ken/formattable")
df <- data.frame(
  id = 1:10,
  name = c("Bob", "Ashley", "James", "David", "Jenny", 
    "Hans", "Leo", "John", "Emily", "Lee"), 
  age = c(28, 27, 30, 28, 29, 29, 27, 27, 31, 30),
  grade = c("C", "A", "A", "C", "B", "B", "B", "A", "C", "C"),
  test1_score = c(8.9, 9.5, 9.6, 8.9, 9.1, 9.3, 9.3, 9.9, 8.5, 8.6),
  test2_score = c(9.1, 9.1, 9.2, 9.1, 8.9, 8.5, 9.2, 9.3, 9.1, 8.8),
  final_score = c(9, 9.3, 9.4, 9, 9, 8.9, 9.25, 9.6, 8.8, 8.7),
  registered = c(TRUE, FALSE, TRUE, FALSE, TRUE, TRUE, TRUE, FALSE, FALSE, FALSE),
  stringsAsFactors = FALSE)
library(formattable)

f <- formattable(df, list(
  age = color_tile("white", "orange"),
  grade = formatter("span",
    style = x ~ ifelse(x == "A", style(color = "green", font.weight = "bold"), NA)),
  test1_score = normalize_bar("pink", 0.2),
  test2_score = normalize_bar("pink", 0.2),
  final_score = formatter("span",
    style = x ~ style(color = ifelse(rank(-x) <= 3, "green", "gray")),
    x ~ sprintf("%.2f (rank: %02d)", x, rank(-x))),
  registered = formatter("span", 
    style = x ~ style(color = ifelse(x, "green", "red")),
    x ~ icontext(ifelse(x, "ok", "remove"), ifelse(x, "Yes", "No")))
))
repr_html.htmlwidget <- function(x, ...){
    htmlfile <- tempfile(fileext = ".html")

    # save the widget to HTML 
    htmlwidgets::saveWidget(
     widget = x, 
     file = htmlfile,
     selfcontained = TRUE
    )

    html_string <- readChar(htmlfile, file.info(htmlfile)$size)
    return(html_string)
}

repr_html.formattable <- function(x){
    repr_html.htmlwidget(as.htmlwidget(x))
}
f
```
-> will show a colored table....

The problem here was that 

```R
x <- print(f)
str(x)
```
shows that x has something... :-(

takluyver:
IIRC, we ran into something that hides output deliberately by making `print.whatever()` do nothing, because for whatever reason it couldn't use the normal visibility mechanism. We were unexpectedly displaying rich reprs in that case, whereas normally there would be no output. So we decided that if there's no plain text output, we should hide all rich reprs as well, to preserve behaviour.

janschulz:
If that's so, we really need to change this here, so that if `print(obj)` returns something we do the repr dance for that returned object... But that means that we have to either remove the `repr_text` call and use print directly (if we can both capture the output and the return object) or add something to the repr_text.default to send back the returned object.

janschulz:
Unfortunatelly, this does not tell what was the problem: https://github.com/IRkernel/IRkernel/commit/7d0dddced70b0dcc69c2353881c550cadf5cadcc

BTW, this would again be solved if we only use repr_text per default and no repr_html... -> https://github.com/IRkernel/IRdisplay/issues/18#issuecomment-200860447 :-)

takluyver:
I think #127 is the relevant issue.

The object being displayed is a data table, for which it does make sense to have rich reprs, so it wouldn't be solved by not doing rich reprs for primitive types.

flying-sheep:
CommManager also uses debug. just reverse the renaming of debug, then all should work

flying-sheep:
hmm, it doesn’t show anything for me!

```
DEBUG: Found full html page: <!DOCTYPE html>
<html>
<head>
<meta charset="utf-8" /> <meta name="viewport" content="width=device-w...
DEBUG: Sending display_data: List of 3
 $ text/plain: chr ""
 $ text/html : chr "<!DOCTYPE html>\n<html>\n<head>\n<meta charset=\"utf-8\" /> <meta name=\"viewport\" content=\"width=device-width, initial-scale"| __truncated__
 $ text/latex: chr "\\begin{tabular}{r|llllllll}\n  & id & name & age & grade & test1\\_score & test2\\_score & final\\_score & registered\\\\\n\\h"| __truncated__
```

flying-sheep:
so if we can’t fix that one, let’s redo the history of [this branch](https://github.com/IRkernel/IRkernel/tree/isolate-full-html) with my fixes and then merge it in.

janschulz:
Not sure about #127, but the https://github.com/IRkernel/IRkernel/commit/7d0dddced70b0dcc69c2353881c550cadf5cadcc is the commit...

janschulz:
I've just built this:

```
 handle_value <- function(obj) {
            data <- namedlist()
            metadata <- namedlist()

            # We have to check the return value of print if it is visible and use the last object
            # which is visible to compute the output from. There are a few packages which use print()
            # to convert itself to something else (and don't output anything to the console) and the
            # returned obj is the printed again.
            print_return <<- list(visible=TRUE, value=obj)
            send_even_if_no_print <- FALSE
            while(returned_obj$visible) {
                data[['text/plain']] <- paste(capture.output(print_return <<- withVisible(print(print_return$value))), collapse = '\n')
                if (returned_obj$visible) repr_even_if_no_print <- TRUE
            }
            obj <- print_return$value

            # data.table has some logic which can't use invisible() and therfore overwrites print to not 
            # return anything
            if (send_even_if_no_print || nchar(data[['text/plain']]) > 0) {
                if (getOption('jupyter.rich_display')) {
                    for (mime in getOption('jupyter.display_mimetypes')) {
                        # Use withCallingHandlers as that shows the inner stacktrace:
                        # https://stackoverflow.com/questions/15282471/get-stack-trace-on-trycatched-error-in-r
                        # the tryCatch is  still needed to prevent the error from showing
                        # up outside further up the stack :-/
                        tryCatch(withCallingHandlers({
                                r <- mime2repr[[mime]](obj)
                                if (!is.null(r)) {
                                    data[[mime]] <- r
                                    # Isolating full html pages (putting them in an iframe)
                                    if (identical(mime, 'text/html')) {
                                        if (grepl("<html.*>", r, ignore.case = TRUE)){
                                            jupyter_debug("Found full html page: %s", strtrim(r, 100))
                                            metadata[[mime]] <- list(isolated = TRUE)
                                        }
                                    }
                                }
                            }, error = handle_display_error),
                            error=function(x){})
                    }
                }
                jupyter_debug("Sending display_data: %s", paste(capture.output(str(data)), collapse = "\n"))
                send_response('display_data', request, 'iopub', list(
                    data = data,
                    metadata = metadata))
            }
        }
```

flying-sheep:
dude. ```` ```r ```.

janschulz:
@flying-sheep Just do it, you have edit right ... Also: sorry :-)

flying-sheep:
haha np, it’s just that i did it so many times already :wink: 

you could cherry-pick 2bd664349ae328727af9a7139e84f989e6f24d6c to get my changes into this PR.

janschulz:
added my stuff and cherry picked your commit... I will do some tests here and then it would be nice if someone of you could do a last pass over the code and merge if everything looks ok.

flying-sheep:
great! please post again once your side is ready

janschulz:
Ok, the earlier commit was borked...

Anyway, I this now works:

* html widgets are now displayed (when adding a `repr_html.htmlwidget`)
* data.table are not displayed when adding a new column (unfortunately also not when it should print, eg a cell which just contains a reference to the data.table )

R/execution.r
```diff
@@ -203,41 +203,48 @@ execute = function(request) {
...
+                                        log_debug("Found full html page: %s", strtrim(r, 100))
+                                        metadata[[mime]] <- list(isolated = TRUE)
                                     }
                                 }
-                            }, error = handle_display_error),
-                            error=function(x){})
-                    }
+                            }
+                        }, error = handle_display_error),
+                        error=function(x){})
```
flying-sheep: code style: spaces around `=`, space between `){`

takluyver:
> data.table ... also not when it should print, eg a cell which just contains a reference to the data.table

Is there any way to avoid that? It seems likely to annoy people.

flying-sheep:
what does “when adding a new column” mean here?

janschulz:
```R
library(data.table)
dat <- data.table(x1=1:10)
dat[, x2 := 1:10] # should not and does not print
dat # does not print anything
print(dat) # prints something
```

IMO the fix for the printing must go into data.table itself: it already has a workaround for knit_print

janschulz:
This is the current `data.table:::print.data.table` function:

```R
function (x, topn = getOption("datatable.print.topn"), nrows = getOption("datatable.print.nrows"), 
    row.names = TRUE, quote = FALSE, ...) 
{
    if (.global$print != "" && address(x) == .global$print) {
        SYS <- sys.calls()
        if (length(SYS) <= 2 || (length(SYS) > 3L && SYS[[length(SYS) - 
            3L]][[1L]] == "knit_print.default")) {
            .global$print = ""
            return(invisible())
        }
    }
    .global$print = ""
    if (!is.numeric(nrows)) 
        nrows = 100L
    if (!is.infinite(nrows)) 
        nrows = as.integer(nrows)
    if (nrows <= 0L) 
        return(invisible())
    if (!is.numeric(topn)) 
        topn = 5L
    topnmiss = missing(topn)
    topn = max(as.integer(topn), 1L)
    if (nrow(x) == 0L) {
        if (length(x) == 0L) 
            cat("Null data.table (0 rows and 0 cols)\n")
        else cat("Empty data.table (0 rows) of ", length(x), 
            " col", if (length(x) > 1L) 
                "s", ": ", paste(head(names(x), 6), collapse = ","), 
            if (ncol(x) > 6) 
                "...", "\n", sep = "")
        return()
    }
    if (topn * 2 < nrow(x) && (nrow(x) > nrows || !topnmiss)) {
        toprint = rbind(head(x, topn), tail(x, topn))
        rn = c(seq_len(topn), seq.int(to = nrow(x), length.out = topn))
        printdots = TRUE
    }
    else {
        toprint = x
        rn = seq_len(nrow(x))
        printdots = FALSE
    }
    toprint = format.data.table(toprint, ...)
    if (isTRUE(row.names)) 
        rownames(toprint) = paste(format(rn, right = TRUE, scientific = FALSE), 
            ":", sep = "")
    else rownames(toprint) = rep.int("", nrow(toprint))
    if (is.null(names(x))) 
        colnames(toprint) = rep("NA", ncol(toprint))
    if (printdots) {
        toprint = rbind(head(toprint, topn), `---` = "", tail(toprint, 
            topn))
        rownames(toprint) = format(rownames(toprint), justify = "right")
        print(toprint, right = TRUE, quote = quote)
        return(invisible())
    }
    if (nrow(toprint) > 20L) 
        toprint = rbind(toprint, matrix(colnames(toprint), nrow = 1))
    print(toprint, right = TRUE, quote = quote)
    invisible()
}
```
I must confess that I've no idea why it doesn't print: AFAIU the code, it should print... Maybe someone of you can propose a fix (or at least tell me what `.global` is :-))

R/execution.r
janschulz: will add a new commit for this... and `s/"/'/g`...

flying-sheep:
uh, i remember something about this. someone explained that in some issue here

janschulz:
Ach damn, now I see that `print(formattable)` doesn't even return a visible value.

janschulz:
So, using the debugger in RStudio (we really want `browser()` support :-/), I think I found out whats going on: `print.formattable` uses  `print` internally again to print, and that triggers rstudios internal stuff: https://github.com/rstudio/rstudio/blob/374557bdaf4423c016f41b6a43ffb2e9854fc862/src/cpp/session/modules/NotebookHtmlWidgets.R#L18-L50

So IMO the solution is here to remove most of the stuff in this PR and implement a `print.htmlwidgets` method, which internally uses display / sends display_data directly instead of relying on the repr_xxx system in handle_value.

flying-sheep:
hey @janschulz: i rebased and squashed your commits in [here](https://github.com/IRkernel/IRkernel/tree/isolate-full-html)

then i fast-forwarded master to have the code style fixes and logging system.

so if you want to continue working on this, you shoud do:

```sh
git fetch origin
git checkout isolate_full_html
git reset --hard origin/isolate-full-html
```

and continue from there.

janschulz:
Argh, the commit messages are all off :-/ (you have code style, I have commit messages :-)) Ok, will make a PR which has all the fixes for the tracebacks and then throw in a commit for a `print.htmlwidgets` in IRDisplay so that we handle that in a manner similar to RStudio.

janschulz:
So, updated with only the relevant commits. This seems to work in my test notebook, but somehow travis does not kick in here?

flying-sheep:
oh yeah, damn, should have used `squash` instead of `fixup`. sorry! but we can get the messages out of the git log of an untampered history.

flying-sheep:
close→reopen→tests pass→merge

janschulz:
Great!

flying-sheep:
indeed! thank you!

flying-sheep:
@janschulz uh, vectors don’t print anymore somehow.

PR #287, opened 2016-03-30
Display htmlwidgets on print

janschulz:
formattable and htmlwidgets expect to be displayed when the "thing" is printed.
For that to work, we need to add a print method which sends the html over the
wire.

We can't use a special repr_html function because `print(obj)` actually does not
print anything and our current system assumes that an empty print indicates that
the rest of the repr functions should also not be shown (which is needed for
packages like data.table which do print magic to work around some issues with
subsetting/assigning new columns not able to return invisible).

```R
df <- data.frame(
  id = 1:10,
  grade = c("C", "A", "A", "C", "B", "B", "B", "A", "C", "C"),
  stringsAsFactors = FALSE)
library(formattable)

f <- formattable(df, list(
  grade = formatter("span",
    style = x ~ ifelse(x == "A", style(color = "green", font.weight = "bold"), NA))
))
f # prints when repr_text calls print
print(f)
```

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -129,6 +129,34 @@ execute = function(request) {
     # shade base::quit
     assign('quit', quit, envir = .GlobalEnv)
     assign('q',    quit, envir = .GlobalEnv)
+
+    # This is needed to get htmlwidgets displayed on print
+    print.htmlwidget <- function(x, ...){
+
+        if (!requireNamespace("htmlwidgets", quietly = TRUE)) {
+            stop("print.htmlwidget called without htmlwidgets loaded")
+        }
+
+        htmlfile <- tempfile(fileext = ".html")
+
+        # save the widget to HTML
+        htmlwidgets::saveWidget(
+            widget = x,
+            file = htmlfile,
+            selfcontained = TRUE
+        )
+
+        html_string <- readChar(htmlfile, file.info(htmlfile)$size)
+        content <- namedlist()
+        metadata <- namedlist()
+        content[['text/html']] <- html_string
+        # self contained -> it's a complete html file which needs to go into
+        # a iframe.
+        metadata[['text/html']] <- list(isolated = TRUE)
+        publish_mimebundle(content, metadata)
+    }
+
+    assign('print.htmlwidget', print.htmlwidget, envir = .GlobalEnv)
     
     send_plot <- function(plotobj) {
         formats <- namedlist()


R/execution.r
```diff
@@ -129,6 +129,27 @@ execute = function(request) {
...
+
+        html_string <- readChar(htmlfile, file.info(htmlfile)$size)
+        content <- namedlist()
+        metadata <- namedlist()
+        content[['text/html']] <- html_string
+        metadata[['text/html']] <- list(isolated = TRUE)
+        publish_mimebundle(content, metadata)
+    }
+
+    assign('print.htmlwidget', print.htmlwidget, envir = .GlobalEnv)
```
janschulz: I suspect that we end up with a few more such functions (quit is already there; cran chooser, debug stuff, password input...). Is there a way to refactor this into a new file? Probably be easiest would be if there is a way to add a new environment so that it is available both in the kernel and for the user and the executer only adds it?

flying-sheep:
ugh. must this all be inline in the kernel? can’t we just have it in a namespace? e.g. the kernel namespace?

we have to make sure it’s earlier in the search path than other `print.htmlwidget` methods, though.

aaand probably the kernel isn’t the right place for this at all. maybe htmlwidget should stop using `print` internally and rely on it being called.

flying-sheep:
also, can you show us why a `repr_html.htmlwidget` isn’t enough?

janschulz:
> also, can you show us why a repr_html.htmlwidget isn’t enough?

See the commit message above :-) `print(formattable)` (and `print.htmlwidget`) does not print anything, so the kernel assumes that nothing should be printed and does not call repr on it (and we can't change this because that would break data.table, which does that as well...

This basically follows the lead of RStudio, which does the same: https://github.com/rstudio/rstudio/blob/374557bdaf4423c016f41b6a43ffb2e9854fc862/src/cpp/session/modules/NotebookHtmlWidgets.R

janschulz:
> ugh. must this all be inline in the kernel? can’t we just have it in a namespace? e.g. the kernel namespace?

I would like to, but have no clue how to do that (see inline comment...) As far as I understand the RStudio code, it uses a special environment...

> aaand probably the kernel isn’t the right place for this at all. maybe htmlwidget should stop using print internally and rely on it being called.

Maybe if you get RStudio to adopt repr for representing the console output :-)

janschulz:
> we can also provide a repr_text.htmlwidget to prevent it from behaving wrong here.

That would actually be a workaround: `repr_text.formattable` and `repr_text.htmlwidgets` which return something (`html widget only visible in when rich displaying is enabled`) and then get a `repr_html.htmlwidget` and `repr_html.formattable` included.

But this means that we have to include a `repr_{html|text}.<packagename>` for all packages which internally use the same pattern as formattable while the single `print.htmlwidget` does the right thing...

flying-sheep:
the choice is still between hacking package-specific behavior into the kernel (which we don’t do yet, and the kernel is already pretty complex/hacky) or handling package-specific displaying in repr, which is built for that.

i’m all for the latter.

R/execution.r
flying-sheep: very good point. we shoud do that.

as you said, the best thing would be an environment that’s always between the .GlobalEnv and all packages in the search path if that’s at all possible.

janschulz: `s/we/you/` as I've no clue how that works :-)

flying-sheep:
i’ll close this and simply implement it in repr.

one problem though: the iframe ends up having `style="…;height:1px"`. why?

janschulz:
> height:1px

I've no clue, but in my case it works as in "I've a colorful dataframe"

flying-sheep:
weird. i cc’d you in a notebook issue about this. the notebook also has your dataframe example riding on the new htmlwidget code in repr

PR #289, opened 2016-04-04
Update URL creation for the case of IPC transport.

craigcitro:
In the case that the kernel manager <--> kernel communication happens via IPC,
there's a slight change in the way the address is formed (`-` instead of `:`
for joining the port). This shows up in the Jupyter code here:
  https://github.com/jupyter/jupyter_client/blob/cf6d354c7e81f6c61879beaf95c1f801486f758c/jupyter_client/connect.py#L442-L445

This change just makes the same tweak in IRkernel.

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -289,10 +289,12 @@ shutdown = function(request) {
 
 initialize = function(connection_file) {
     connection_info <<- fromJSON(connection_file)
+    stopifnot(connection_info$transport %in% c('tcp', 'ipc'))
     
     url <- paste0(connection_info$transport, '://', connection_info$ip)
     url_with_port <- function(port_name) {
-        paste0(url, ':', connection_info[[port_name]])
+        sep <- switch(connection_info$transport, tcp = ':', ipc = '-')
+        paste0(url, sep, connection_info[[port_name]])
     }
     
     # ZMQ Socket setup


flying-sheep:
thanks! can you please reduce the amount of lines and use `<-` for assignment?

e.g.

```r
sep <- switch(connection_info$transport, tcp = ':', '-')
```

or

```r
sep <- if (connection_info$transport == 'tcp') ':' else '-'
```

craigcitro:
Done.

flying-sheep:
good idea, but now if there’s an unknown transport, it will fall back to `NULL`, which is left out in the following url like `paste0(url, port)`

maybe we should do this?

```r
sep <- switch(connection_info$transport, tcp = ':', ipc = '-',
    stop('Unknown transport ', connection_info$transport, ' in connection file'))
```

(also please don’t use `''` or ``` `` ``` for named parameters unless necessary)

craigcitro:
I ended up tweaking just slightly -- since the transport doesn't change in the course of the function, I just checked as we entered the function. I also used `stopifnot` -- it still gives a readable error, and the only way someone is likely to hit this is if they're manually constructing a connection file, which seems pretty rare.

I have no idea why I had `'tcp'` instead of just `tcp`. :wink:

flying-sheep:
thank you!

the check isn’t local so new protocols will have to be added in two places, but as you said that’s all unlikely, and i was pedantic enough already :wink:

PR #292, opened 2016-04-08
Save plot size in each recordedplot and use it to build images

janschulz:
This fixes the problem that the size of a plot would be taken from the options
when a new plot was created or when the cell was finished and not when the
original plot was created. This resulted in problems when you set the plot size
between two plots in the same cell or more specificly, you would have to set the
plot size *after* you created a plot (and before the cell was finished or a new
plot was created.

No we save the plot size at the time the plot is created and use that size to
build the images.

```
set_plot_size <- function(width = 7, height = 7){
    options(repr.plot.width = width)
    options(repr.plot.height = height)
}
set_plot_size(3,1) # ignored before, now taken for the first plot
qplot(1:10,1:10)
set_plot_size(3,3) # before for the first plot, now for the second
qplot(1:10,1:10)
set_plot_size(1,3) # before taken for the second plot, now only for following plots
```

Closes: https://github.com/IRkernel/repr/issues/42

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -135,7 +135,7 @@ execute = function(request) {
         metadata <- namedlist()
         for (mime in getOption('jupyter.plot_mimetypes')) {
             tryCatch({
-                formats[[mime]] <- mime2repr[[mime]](plotobj)
+                formats[[mime]] <- mime2repr[[mime]](plotobj, attr(plotobj, '.irkernel_width'), attr(plotobj, '.irkernel_height'))
             }, error = handle_error)
             # Isolating SVGs (putting them in an iframe) avoids strange
             # interactions with CSS on the page.
@@ -221,7 +221,7 @@ execute = function(request) {
                     }
                 }
             }
-            log_debug("Sending display_data: %s", paste(capture.output(str(data)), collapse = "\n"))
+            log_debug('Sending display_data: %s', paste(capture.output(str(data)), collapse = "\n"))
             send_response('display_data', request, 'iopub', list(
                 data = data,
                 metadata = metadata))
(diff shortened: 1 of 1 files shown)


flying-sheep:
Great, thanks, only one small thing:

R/execution.r
```diff
@@ -237,6 +237,11 @@ execute = function(request) {
             if (!plot_builds_upon(last_recorded_plot, plotobj)) {
                 send_plot(last_recorded_plot)
             }
+            # need to be set here to capture the size and have it available
+            # when the plot is sent
+            # 7 is the default in repr...
+            attr(plotobj, '.irkernel_width') <- getOption("repr.plot.width", 7)
+            attr(plotobj, '.irkernel_height') <- getOption("repr.plot.height", 7)
```
flying-sheep: i don’t like code repeating itself.

IRkernel depends on repr, and repr sets the option on load, so the option will be set unless explicitly unset by an user. please remove the defaults from the getOption calls and the “7 is the default in repr...” comment

janschulz: > also single quotes

damn...

I don't like code crashing when an option is deleted by a user, so I think we should add a default here. Let's add it as a TODO, change repr to export the defaults, wait a few month and then change irkernel to use the defaults.

janschulz:
Added a TODO to remove it when repr changes and s/"/'/...

R/execution.r
flying-sheep: why wait? i think you can expect IRkernel dev versions to not work without the newest repr

flying-sheep:
OK, it’s now in repr: IRkernel/repr@63e227a19bef5d920e320aeb630d92d5fec4f310

we can now directly use it and get rid of comment, TODO & magic numbers

flying-sheep:
Merged via 48ac388967115682b820777eaeb884f689e471fb

janschulz:
> we can now directly use it and get rid of comment, TODO & magic numbers

No: this will crash if someone updates irkernel but not repr...

flying-sheep:
done: 752ced6df9d093f0c5439f618346ab40813d0e18

PR #293, opened 2016-04-10
Log2file

janschulz:
This is mainly a byproduct of https://github.com/IRkernel/repr/issues/21, where I wanted to see where the wrong chars come from and needed to log to a file.

### DIFF
--- a/R/logging.r
+++ b/R/logging.r
@@ -2,24 +2,58 @@
 # Per default, only error messages are shown
 # levels: 3 = DEBUG, 2 = INFO/MSG, 1 = ERROR
 
-# Handle for stderr to even log to the console when the stderr() points to a
-# sink'ed connection...
-stderror <- stderr()
-
 log_debug <- function(...) {
-    if (getOption('jupyter.log_level') >= 3L) {
-        cat('DEBUG: ', sprintf(...), '\n', sep = '', file = stderror)
+    if (isTRUE(getOption('jupyter.log_level') >= 3L)) {
+        log_msg('DEBUG', sprintf(...))
     }
 }
 
 log_info <- function(...) {
-    if (getOption('jupyter.log_level') >= 2L) {
-        cat('INFO: ', sprintf(...), '\n', sep = '', file = stderror)
+    if (isTRUE(getOption('jupyter.log_level') >= 2L)) {
+        log_msg('INFO', sprintf(...))
     }
 }
 
 log_error <- function(...) {
-    if (getOption('jupyter.log_level') >= 1L) {
-        cat('INFO: ', sprintf(...), '\n', sep = '', file = stderror)
+    if (isTRUE(getOption('jupyter.log_level') >= 1L)) {
+        log_msg('ERROR', sprintf(...))
     }
+}
+
+log_msg <- function(lvl, msg){
+        log_msg_stderror(lvl, msg)
+        log_msg_logfile(lvl, msg)
+}
+
+
+# Handle for stderr to even log to the console when the stderr() points to a
+# sink'ed connection...
+.stderror <- stderr()
+
+log_msg_stderror <- function(lvl, msg){
+    cat(sprintf('%s: %s\n', lvl, msg) , file = .stderror)
+}
+
+.is_changed_logfile <- local({
+	old_logfile <- ""
+	function(logfile) {
+		if (old_logfile != logfile) {
+			old_logfile <<- logfile
+			TRUE
+		} else {
+			FALSE
+		}
+	}
+})
+
+log_msg_logfile <- function(lvl, msg){
+	if (!is.null(getOption('jupyter.logfile'))) {
+		cur_logfile <- getOption('jupyter.logfile')
+		if (.is_changed_logfile(cur_logfile)) {
+			log_msg_stderror('INFO', sprintf('Logging to %s', cur_logfile))
+		}
+		log_con <- file(cur_logfile, open='ab')
+		writeBin(charToRaw(sprintf('%s %s: %s\n', format(Sys.time()), lvl, msg)), log_con, endian='little')
+		close(log_con)
+	}
 }
--- a/R/logging.r
+++ b/R/logging.r
@@ -46,14 +46,14 @@ log_msg_stderror <- function(lvl, msg){
 	}
 })
 
-log_msg_logfile <- function(lvl, msg){
+log_msg_logfile <- function(lvl, msg) {
 	if (!is.null(getOption('jupyter.logfile'))) {
 		cur_logfile <- getOption('jupyter.logfile')
 		if (.is_changed_logfile(cur_logfile)) {
 			log_msg_stderror('INFO', sprintf('Logging to %s', cur_logfile))
 		}
 		log_con <- file(cur_logfile, open='ab')
-		writeBin(charToRaw(sprintf('%s %s: %s\n', format(Sys.time()), lvl, msg)), log_con, endian='little')
+		writeBin(charToRaw(sprintf('%s %s: %s\n', format(Sys.time()), lvl, msg)), log_con, endian = 'little')
 		close(log_con)
 	}
 }
--- a/R/options.r
+++ b/R/options.r
@@ -12,8 +12,7 @@ getenv_default <- function(varname, default) {
 #' @export
 jupyter_option_defaults <- list(
     jupyter.log_level = as.integer(getenv_default('JUPYTER_LOG_LEVEL', 1L)),
-    jupyter.log_to_file = FALSE,
-    jupyter.logfile = getenv_default('JUPYTER_LOGFILE', 'irkernel.log'),
+    jupyter.logfile = getenv_default('JUPYTER_LOGFILE', NULL),
     jupyter.pager_classes = c(
         'help_files_with_topic'),
     jupyter.plot_mimetypes = c(
--- a/R/options.r
+++ b/R/options.r
@@ -12,6 +12,8 @@ getenv_default <- function(varname, default) {
 #' @export
 jupyter_option_defaults <- list(
     jupyter.log_level = as.integer(getenv_default('JUPYTER_LOG_LEVEL', 1L)),
+    jupyter.log_to_file = FALSE,
+    jupyter.logfile = getenv_default('JUPYTER_LOGFILE', 'irkernel.log'),
     jupyter.pager_classes = c(
         'help_files_with_topic'),
     jupyter.plot_mimetypes = c(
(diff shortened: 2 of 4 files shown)


janschulz:
This should get a better default location for the log, but for that I would need to know where to put logfiles... @takluyver Is there some "default location", where all kernels should log to? the ipykernel still uses the old ipython profile support, so that is not gonna work for us...

R/logging.r
```diff
@@ -7,19 +7,53 @@
 stderror <- stderr()
 
 log_debug <- function(...) {
-    if (getOption('jupyter.log_level') >= 3L) {
-        cat('DEBUG: ', sprintf(...), '\n', sep = '', file = stderror)
+    if (isTRUE(getOption('jupyter.log_level') >= 3L)) {
```
janschulz: Is added `isTRUE` to guard against the case where a user removes an option...

flying-sheep:
i’d prefer to use `jupyter.logfile` and test `if (!is.null(getOption('jupyter.logfile')))`. it has exactly the same amout of information as using two options.

R/logging.r
```diff
@@ -7,19 +7,53 @@
...
+        log_msg_stderror('DEBUG', sprintf(...))
+        log_msg_logfile('DEBUG', sprintf(...))
```
flying-sheep: i’d design it like this:

```r
log_debug <- function(...) {
	if (isTRUE(getOption('jupyter.log_level') >= 3L)) {
		log_msg('DEBUG', sprintf(...))
	}
}

log_msg <- function(lvl, msg){
	log_msg_stderror(lvl, msg)
	if (isTRUE(getOption('jupyter.log_to_file'))) {
		log_msg_logfile(lvl, msg)
	}
}

log_msg_stderror <- ...
log_msg_logfile <- ...  #without the if()
```

flying-sheep:
all those design considerations nonwithstanding, i’m actually against enhancing our logging, and for switching to a dedicated logging system.

1. playing together with other things that log is important
2. this is too complex and too different a concern to manage inside of the repo
3. we get too many logging-related config options that whould be rather global

obviously all of this helps with debugging, but i don’t want us to live in our own NIH-fueled wheel-reinventing bubble.

we obviously need a solution that has per-package logging so that we can debug only the kernel without seeing all the crap packages log.

@janschulz are you OK with closing this and deciding on a dedicated solution in #119 this week

R/logging.r
janschulz: tried that, but then wanted to send the filename to the normal log and didn't want to get into recursion and so split both calls...

janschulz:
I'm not so sure: I agree that we should go to a proper lib, but unless someone does that, it would be nice to have a way to get properly encoded logs from the kernel (the log in the notebook windows is garbled wrt unicode... :-()

janschulz:
> i’d prefer to use jupyter.logfile and test if (!is.null(getOption('jupyter.logfile'))). it has exactly the same amout of information as using two options.

ok, sounds reasonable...

janschulz:
Whatever we do merge the log_to_file part: the other commit of adding more logging calls in the real code should be merged...

janschulz:
> ok, sounds reasonable...

Ok, I take that back: IMO it should be "enable it and get a logfile wherever jupyter logs to, too". But that's probably better done in a later version...

flying-sheep:
> Ok, I take that back: IMO it should be "enable it and get a logfile wherever jupyter logs to, too". But that's probably better done in a later version...

i don’t get it.

my semantics for a single log file option would optimally be:

1. if NULL, don’t log to a file
2. if a character vector, log to (each of the) file(s) specified
3. if something else, throw an error

---

but the above behavior is for IRkernel, not our logging system.

a proper logging system should allow registering arbitrarily many loggers: to files, to databases, to, stderr, …

janschulz:
My semantics are based on the one in ipython: if you enable logging in the ipyparallel engines, AFAIK it logs to `%USERPROFILE%\.ipython\profile_default\log`. Not sure how you let the notebook log to a file (probably [not at all yet](https://github.com/ipython/traitlets/blob/a2543e109990e3e5784a7c616a5517fc99562b60/traitlets/config/application.py#L208)). Julia also seem to [only log to stdout]( https://github.com/JuliaLang/IJulia.jl/blob/a675ac5c5b08bf75a2f4d2f3b80035252a4af3c1/src/stdio.jl#L15).

R/logging.r
flying-sheep: eh, you can still used the individual ones, i just wrapped them into `log_msg`.

flying-sheep:
hmm, OK. i think jupyter specific logging options won’t hurt. bonus points if we can configure the future real logging system according to them.

about the paths, you should check `jupyter --help` or `jupyter --paths`:

* `jupyter --runtime-dir` is sth like `/run/user/1000/jupyter`
* `jupyter --config-dir` is (currently/per default) `~/.jupyter`
* `jupyter --data-dir` is `$XDG_DATA_DIR/jupyter` aka `~/.local/share/jupyter`

so if anything of those, the last one should be used. the first one isn’t available after kernel shutdown, and the second one is the *config* dir, not “all kinds of stuff dir”

R/logging.r
janschulz: Ok, no I got it.... Update is on the way...

janschulz:
Checking this is not gonna work: You don't need to have jupyter stuff in path (e.g. only installed in a conda environment)...

flying-sheep:
well, then there’s no possible way to deduce the correct jupyter directory.

but you know what: the correct system log directory is

* `/var/log/irkernel/` on linux and OSX,  and
* `%ProgramData%\IRkernel\logs\` on windows

the correct user log dir is

* `${XDG_CACHE_HOME-$HOME/.cache}/irkernel/logs` on linux
* `%ProgramData%\Roaming\IRkernel\logs\` on windows, and
* `~/Library/Logs/IRkernel` on OSX.

if you want you can provide a function:

```r
get_log_dir() {
	logfile <- getOption('jupyter.logfile')
	if (is.null(logfile) || identical(logfile, FALSE)) {
		NULL
	} else if (is.character(logfile)) {
		logfile
	} else if (isTRUE(logfile)) {
		# logic to get OS specific log dir
	} else stop(sprintf('Option %s of unknown type %s instead of logical, character, or NULL', sQuote('jupyter.logfile'), class(logfile)))
}
```

janschulz:
@takluyver Do you know of any recommendation for kernels where to place logfiles?

flying-sheep:
the above are OS-level recommendations on where *everything* should put logfiles.

takluyver:
Sorry for the delay on this. At the dev meeting, we did get some time to discuss our 'kernel nanny' plan, which will include opt-in OS-level output capturing. As part of that, the kernel will get an environment variable with a path it can log into for debugging the kernel itself. Where this goes will be up to the server - it might be written to a log file, or it might be sent straight to the terminal where you started the notebook server (the ipython kernel already logs to the terminal).

Now we just need to implement that proposal! ;-)

flying-sheep:
hmm, so the logging config is on the server side… and the env variable points to a FIFO read by the server or what?

takluyver:
We haven't decided where we'll send logs yet, and it will probably be configurable on the server side. The idea is that the env var points to something that the kernel can `open()` and write to, and the kernel shouldn't need to worry about exactly what that is. It could be a regular file, a FIFO, or even a reference to the tty the notebook server is running in, on platforms where that's doable.

flying-sheep:
sounds nice!

janschulz:
@takluyver Any issues I can subscribe to, to follow along?

janschulz:
So, after hearing this, my vote goes to merging this as is and then change/remove stuff in the next iteration, if that's needed..

takluyver:
There's not currently any issue or anything about this, but I'll try to remember to keep you posted about it.

R/options.r
```diff
@@ -12,6 +12,7 @@ getenv_default <- function(varname, default) {
 #' @export
 jupyter_option_defaults <- list(
     jupyter.log_level = as.integer(getenv_default('JUPYTER_LOG_LEVEL', 1L)),
+    jupyter.logfile = getenv_default('JUPYTER_LOGFILE'),
```
flying-sheep: won’t this say “missing argument, no default”?

maybe we should change the definition of `getenv_default` to `function(varname, default = NULL)`?

flying-sheep:
> So, after hearing this, my vote goes to merging this as is and then change/remove stuff in the next iteration, if that's needed..

yeah, that’s good!

we can always change the `jupyter.logfile` to have more possible values than an explicit file path or `NULL` (add `TRUE` to mean “jupyter’s location”)

flying-sheep:
OK, outstanding points:

* extract `send_error_msg` and `is_silent` (or so)
* does `getenv_default` have to be fixed for your usage of it?
* some remaining `){` after functions and some `"`s

janschulz:
I fixed the style fixes and the default value.

> extract send_error_msg and is_silent (or so)

You mean as additional class methods? I still don't get how these functions would have access to the request context.

flying-sheep:
not at all:

```r
	send_error_msg = function(request, msg) {
		if (!is_silent(request)) return()
		...
	},
	is_silent = function(request) {...},
}

maybe we can come up with something better at another time, but this is better than inline

janschulz:
So, what is worse: redefining all functions on each request or doing two object lookups for each `if (!is_silent(request)) ...`? Probably the function definitions?

-> Will refactor...

flying-sheep:
you mean performance-wise? i don’t care, none of this should matter. also making performance arguments before measuring is really thin ice to tread on.

the point is to put as much *declarative* code as possible to keep the *imperative* part small (don’t define functions at runtime if not necessary; don’t put too much spaghetti code into one function)

janschulz:
So, promoted these two functions to methods...

janschulz:
Big refactoring of execute incoming...

janschulz:
Ok, this has now a much shorter `Executer$execute()` method... @flying-sheep Please have a look, this is my first work with the such ref classes and it took a few rounds until I had it working...

janschulz:
If this is too high risk, it would be good to get everythin up to 75aea9d merged. I can then add another PR just for the refactoring...

flying-sheep:
wow cool, this is really great!

i actually just wanted you to not add to the pile but instead you fixed it :laughing: 

i’ll test it right away

flying-sheep:
works great thanks!

i only found out that my getenv_default stuff doesn’t work (apparently it’s executed on package build time or something)

i’ll merge and fix this.

flying-sheep:
ok, done: b11f2593b0ec7c946df1f58a9ec147d3b8feed8a

PR #296, opened 2016-04-13
Warn unicode escapes

janschulz:
Warn when a usere inputs a unicode char which would be escaped before execution...

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -256,6 +256,25 @@ execute = function(request) {
     interrupted <<- FALSE
     last_recorded_plot <<- NULL
     log_debug('Executing code: %s', request$content$code)
+    
+    # Workaround to warn user when code contains potential problematic code
+    # https://github.com/IRkernel/repr/issues/28#issuecomment-208810772
+    # See https://github.com/hadley/evaluate/issues/66
+    if(.Platform$OS.type == 'windows') {
+        # strip whitespace, because trailing newlines would trip the test...
+        code <- gsub('^\\s+|\\s+$', '', request$content$code)
+        real_len <- nchar(code)
+        r_len <- nchar(paste(capture.output(cat(code)), collapse = '\n'))
+        if (real_len != r_len){
+            msg = c('Your code contains a unicode char which cannot be displayed in your ',
+                    'current locale and R will silently convert it to an escaped form when the ',
+                    'R kernel executes this code. This can lead to subtle errors if you use ',
+                    'such chars to do comparisons. For more information, please see ',
+                    'https://github.com/IRkernel/repr/wiki/Problems-with-unicode-on-windows')
+            send_error_msg(paste(msg, collapse = '\n'))
+        }
+    }
+
     tryCatch(
         evaluate(
             request$content$code,


flying-sheep:
Super nice, especially the wiki page!

https://github.com/IRkernel/repr/wiki/Problems-with-unicode-on-windows

R/execution.r
```diff
@@ -133,6 +134,14 @@ execute = function(request) {
     assign('quit', quit, envir = .GlobalEnv)
     assign('q',    quit, envir = .GlobalEnv)
     
+    send_error_msg <- function(msg){
```
flying-sheep: can you put this as method into the executor (and add a space here)?

maybe like `send_error_msg(request, msg)`? `if (!silent)` would have o go outside of this then though, …

i prefer not to create too many functions in procedural pieces of code.

R/execution.r
```diff
@@ -241,6 +246,23 @@ execute = function(request) {
     interrupted <<- FALSE
     last_recorded_plot <<- NULL
     
+    # Workaround to warn user when code contains potential problematic code
+    # https://github.com/IRkernel/repr/issues/28#issuecomment-208810772
+    # See https://github.com/hadley/evaluate/issues/66
+    if(.Platform$OS.type == "windows") {
```
flying-sheep: " → '

(more below)

R/execution.r
```diff
@@ -241,6 +246,23 @@ execute = function(request) {
...
+        # strip whitespace, because trailing newlines would trip the test...
+        code <- gsub("^\\s+|\\s+$", "", request$content$code)
+        real_len <- nchar(code)
+        r_len <- nchar(paste(capture.output(cat(code)), collapse="\n"))
```
flying-sheep: " → ' again and spaces (also below)

R/execution.r
```diff
@@ -241,6 +246,23 @@ execute = function(request) {
...
+        if (real_len != r_len){
+            msg = c("Your code contains a unicode char which cannot be displayed in your ",
```
flying-sheep: maybe extract to top level as sth. like `win_unicode_warning` and just do

```r
win_unicode_warning <- "Blah
blah blah
blah"
```

R/execution.r
janschulz: I don't think it makes sense to put that anywhere else: if the `silent` is outside the function, it's simple a `send_response` call and then I could simple use that...

R/execution.r
janschulz: damn it... I really hate Rs style on quotes...

R/execution.r
janschulz: ... and spaces... who puts spaces around `=` in function calls?

janschulz:
Will fix code style...

Re code reorg and promoting some to toplevel: I'm not sure how much time is used to redeclare the functions on each execution, but there is a lot of function declaration going on in the execution of code. If it makes a difference, there are a lot of functions which could be declared somewhere else (e.g. as `something(code, silent)` instead of using silent from the outside scope...) or by using a `context` object which the handler could request form an environment?

R/execution.r
flying-sheep: certainly not python :laughing:

but yeah, in R that’s how you do it.

R/execution.r
flying-sheep: It’s a send_response call with most fields filled in. I always create small helpers like that to reduce complexity

flying-sheep:
I don’t think it makes a difference, it’s just horrible spahghetti code (no offense, i added my share of spaghetti on top)

i think we should extract all declarations and put them extra. that can easily be done by adding all variables they grab from scope (including the executor itself) as parameters.

but here maybe you can simply do `if (!silent) send_error_msg(response, msg)`

janschulz:
As far as I understand it, to extract the send_error_msg, I would have to get access to the `send_response` function, but that is only available while in the executor. So I can't move it outside the executer class?

flying-sheep:
yes, therefore we should make it a method of the executor, i think

flying-sheep:
maybe add the methods:

```r
request_is_silent = function(request) request$content$silent,
send_error_msg = function(request, msg) {
    if (!request_is_silent(request)) {
            send_response('stream', request, 'iopub', list(
                name = 'stderr',
                text = msg))
        }
    }
},
```

or so.

janschulz:
IMO a lot of functions could be extracted if there would be a way to get the current request via an methods.

janschulz:
Updated on top of master and repushed

PR #302, opened 2016-04-21
readme: add rzmq dependancy from CRAN

davharris:


janschulz:
Actually, you shouldn't need to install that: rzmq is available from our github site, and should be installed automatically (and rzmq on cran has no win package available, which we have).

-> What was your problem?

[Please also note that our current master depends on a different zmq lib, which has windows packages available on CRAN; we just haven't done any releases for that :-/]

davharris:
Thanks for the quick response.

As for my problem, I tried so many things to fix this in the past hour that I'm not entirely sure, but I think I identified it as a failure to grab zmq.hpp and put it in the right place.  I ultimately just ended up using `wget` to grab the individual file from `https://raw.githubusercontent.com/zeromq/cppzmq/master/zmq.hpp`.  Once I did that, I think my remaining problems all involved undoing various other things I'd attempted to "fix" the problem.

In case it's helpful, here's my `sessionInfo()`, now that everything is working.  Thanks again.

```
R version 3.2.4 (2016-03-10)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: OS X 10.11.4 (El Capitan)

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] magrittr_1.5    IRdisplay_0.3   tools_3.2.4     base64enc_0.1-3
 [5] uuid_0.1-2      stringi_1.0-1   rzmq_0.7.8      IRkernel_0.5   
 [9] jsonlite_0.9.19 stringr_1.0.0   digest_0.6.9    repr_0.4       
[13] evaluate_0.8.3 
```

janschulz:
IMO, the easiest is using this:

```
library("devtools")
install_github("IRkernel/repr")
install_github("IRkernel/IDdisplay")
install.packages("pbdZMQ") # maybe -> IRkernel should install it as a dependency...
install_github("IRkernel/IRkernel") # this should actually install pbdZMQ
```

flying-sheep:
pbdZMQ is listed

Issue #311, opened 2016-05-05
Kernel status messages should be sent before & after each shell request


takluyver:
We currently only send them on execution. I think the spec wasn't clear when we first wrote the code, but this was decided in message spec version 5:
http://jupyter-client.readthedocs.io/en/latest/messaging.html#kernel-status

This would be a relatively easy task if someone is looking to start contributing.

From jupyter/jupyter-js-services#139

janschulz:
@parente, @takluyver Isn't this closed with the merge of #312?

takluyver:
Yep, thanks.

rwhorman:
@takluyver @janschulz  any outlook on a new release incorporating this fix?  we are using jupyter-js-services against R kernels and this makes it rather troublesome to use.

janschulz:
IMO we could send IRdisplay to CRAN and then IRKernel.

dockter-84:
Yes,  I'm interested in incorporating the fix, too.

rwhorman:
@janschulz seems good. so how does this progress to closing on a new release? need to vote somewhere and slip bitcoins to somebody?  ;-)

rwhorman:
@flying-sheep any comment on getting a release?

flying-sheep:
i seem to be a bit inactive in te last weeks, right? :neutral_face: 

well, @janschulz has created milestones for both the kernel and IRdisplay. i’ll see if i get IRdisplay submitted to CRAN this weekend

rwhorman:
@flying-sheep  how did things work out on the weekend? Are we rolling?

flying-sheep:
sorry, no

btw.: i’d like to get IRkernel/IRdisplay#19 fixed first.

rwhorman:
@flying-sheep @janschulz  so now we're waiting on this one? IRkernel/IRdisplay#20

flying-sheep:
i think we’re basically waiting on IRkernel/IRdisplay#33 and IRkernel/IRkernel#368.

it’s important to get this nailed down before we publish (make the packages’ responsibility clear)

PR #312, opened 2016-05-05
Add busy/idle around all shell replies

parente:
For #311

This is done totally by eyeballing it so far. I need to get
a dev environment setup to test it.

I see a test target in the Makefile but I don't see where
tests are defined. I wouldn't be surprised if I'm missing them.
(I'm very much an R newbie.)

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -205,8 +205,6 @@ handle_warning = function(o) {
 },
 
 execute = function(request) {
-    send_response('status', request, 'iopub', list(
-        execution_state = 'busy'))
     send_response('execute_input', request, 'iopub', list(
         code = request$content$code,
         execution_count = execution_count))
@@ -287,8 +285,6 @@ execute = function(request) {
     if ((!is.silent()) & (!is.null(last_recorded_plot))) {
         send_plot(last_recorded_plot)
     }
-    
-    send_response('status', request, 'iopub', list(execution_state = 'idle'))
     
     if (interrupted) {
         reply_content <- list(
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -119,6 +119,11 @@ handle_shell = function() {
     
     parts <- zmq.recv.multipart(sockets$shell, unserialize = FALSE)
     msg <- wire_to_msg(parts)
+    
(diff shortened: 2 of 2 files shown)


takluyver:
I think we could do it in `handle_shell()`, so the code is only needed in one place. Unless I'm missing something.

parente:
If we do it wholesale in `handle_shell`, it slightly changes the order of messages for `execute_request` because now the idle status will come after the `execute_reply` instead of before it like it does now.

Also, unless a special check is added to `handle_shell` for the shutdown case, and busy status message will be sent, but an idle will mostly likely not since there's a quit in the shutdown method. It's probably not a big deal, but perhaps that will leave some UIs in a weird state?

I'm totally willing to switch to doing it in one place. Just let me know if either of the above is a real issue or not.

takluyver:
> If we do it wholesale in handle_shell, it slightly changes the order of messages for execute_request because now the idle status will come after the execute_reply instead of before it like it does now.

They go on different channels, so the order isn't reliable anyway. :-)

> Also, unless a special check is added to handle_shell for the shutdown case, and busy status message will be sent, but an idle will mostly likely not since there's a quit in the shutdown method. It's probably not a big deal, but perhaps that will leave some UIs in a weird state?

Good point. What do other kernels do with status on shutdown_request?

parente:
> They go on different channels, so the order isn't reliable anyway. :-)

Ah, right. I was dwelling on the websocket where it's all serialized to one stream and should have been thinking about zmq.

> Good point. What do other kernels do with status on shutdown_request?

ipykernel sends busy/idle around shutdown and all other messages (https://github.com/ipython/ipykernel/blob/bc4fe8916abb28519adc9214abdb088afa14c323/ipykernel/kernelbase.py#L206). And it leads to the stuck indicator in notebook. :)

screen shot 2016-05-05 at 9 37 18 am

That said, I'm convinced we should replicate what ipykernel is doing and just wrap busy/idle around everything in one place. I'll update the PR.

parente:
Should be good to go.

takluyver:
Thanks @parente

PR #314, opened 2016-05-06
Fixes for R CMD check

janschulz:
This fixes some R CMD check problems and also adresses two errors which were showing when using `jupyter qtconsole --kernel ir33`

Also fixes the travis tests and implements test on all 3 major R versions and with different dependencies.

Closes: #321
Closes: #318


man/IRkernel.Rd
```diff
@@ -9,9 +9,7 @@
 \title{An R kernel for Jupyter.}
 \format{An object of class \code{list} of length 5.}
 \usage{
-options(jupyter.* = ...)
-getOption('jupyter.*')
-jupyter_option_defaults$jupyter.*
+jupyter_option_defaults
```
janschulz: Not sure if there is a better way... Probably with examples, but the code must be valid R, so the `jupyter.*` syntax is out...

man/namedlist.Rd
```diff
@@ -0,0 +1,12 @@
+% Generated by roxygen2: do not edit by hand
+% Please edit documentation in R/execution.r
```
takluyver: Again,  I'm confused as to why this is getting docs.

janschulz:
WTF: 

> The command "curl -Lo /tmp/pandoc-1.15.2-1-amd64.deb https://github.com/jgm/pandoc/releases/download/1.15.2/pandoc-1.15.2-1-amd64.deb" failed and exited with 7 during .

flying-sheep:
such things happen. let’s re-trigger this

R/execution.r
```diff
@@ -272,7 +272,7 @@ execute = function(request) {
             output_handler = oh,
             stop_on_error = 1L),
         interrupt = function(cond) interrupted <<- TRUE,
-        error = handle_error) # evaluate does not catch errors in parsing
+        error = .self$handle_error) # evaluate does not catch errors in parsing
```
janschulz: I really wonder why this only happened in the qtconsole but not the console

flying-sheep: what happened?

R/execution.r
```diff
@@ -114,7 +115,7 @@ quit = function(save = 'default', status = 0, runLast = TRUE) {
         if (!is.null(.GlobalEnv$.Last.sys)) .GlobalEnv$.Last.sys()
     }
     if (save) NULL  # TODO: actually save history
-    payload <<- c(payload, list(list(source = 'ask_exit')))
+    payload <<- c(payload, list(list(source = 'ask_exit', keepkernel = FALSE)))
```
flying-sheep: keepkernel?

R/kernel.r
```diff
@@ -362,9 +362,9 @@ run = function() {
 )
 
 comm_manager_env <- new.env()
-#' Get global CommManager instance
+#' Get global Comm_Manager instance
```
flying-sheep: oh, yes, it’s really called that. missed that one.

that isn’t a naming convention i’ve ever seen in my life. we should rename it to `CommManager`

R/options.r
```diff
@@ -1,7 +1,5 @@
 #' @usage
-#' options(jupyter.* = ...)
-#' getOption('jupyter.*')
-#' jupyter_option_defaults$jupyter.*
+#' jupyter_option_defaults
```
flying-sheep: always sad when automated tests make things actually *worse*.

we should maybe just pick one option that we set and get as an example instead of removing the examples.

man/IRkernel.Rd
flying-sheep: always sad when automated checks make things actually *worse*.

we should maybe just pick one option that we set and get as an example instead of removing the examples.

R/execution.r
janschulz: See link in the commit message. Seems to be an addition?

R/execution.r
janschulz: Before this change, `jupyter console --kernel ir32` (or ir3.3) worked, but `jupyter qtconsole --kernel  ir32` raised an error: https://github.com/IRkernel/IRkernel/issues/319#issuecomment-218404471

man/IRkernel.Rd
janschulz: Looking at the manual on cran, I think we should rework out docstrings anyway... It's a bit sad that in repr none of the actual `repr_html`, `repr_latex`... are actually in the table of content of the PDF :-(

R/kernel.r
janschulz: Agreed, but please after I have this dam travis thing sorted out :-)

.travis.yml
```diff
@@ -2,15 +2,26 @@ language: r
...
   - IRkernel/IRdisplay
 
-addons:
-  apt:
-    packages:
-      - python3
-      - python3-setuptools
 
 script:
   - source .travis-build.sh
```
janschulz: @flying-sheep Why is that sourced and not executed?

janschulz:
Yay, finally it's green... 

3min test times for individual jobs, 6 min overall because we have 6 jobs which is more than the 5 default concurrent ones...

7min for a job which has to compile all dependencies...

-> Please merge if satisfied...

.travis.yml
flying-sheep: i think it’s executed as one command elsewise, instead of as multiple

R/kernel.r
flying-sheep: sure, random breakage for no reason isn’t funny.

R/execution.r
flying-sheep: well as long as it works ¯\\_(ツ)_/¯

flying-sheep:
great stuff! i hope in the long run we can install python in a more lightweight way (e.g. ubuntu package sources)

R/execution.r
janschulz: "now" :-)

.travis.yml
janschulz: Hm, having it executed would be better because than you could just `#! bash -e` instead of the `set -e ... set +e` dance..

There is BTW one big reason to put it all into travis.yml: the commands are echo'ed and you see what command had the error :-)

janschulz:
I think conda is the easiest... I think in the python images, there is even a conda install already...

Issue #301, opened 2016-04-19
Add a way to contibute our own functions to the execution environment


janschulz:
It would be nice if the environments for the kernel and the environment for the execution are separated (from my perspective that looks to be the case) and if there is a simple "entry point" how to add functions and objects into the execution environment:

E.g. if one could do this:

```r
# in "shadows.r"
.add_to_execution_env("vignette", function(...) {...})
.add_to_execution_env("browser", function(...) {...})
.add_to_execution_env("view", function(...)
	stop("View is not (yet) supported in jupyter"))
```

This is currently over my head, maybe @flying-sheep could give me some pointers what would be needed here?

takluyver:
We initially created an environment for user code to execute in, but it turns out that a lot of R stuff assumes that user code is running the global environment, so eventually we had to use the global environment for running user code.

flying-sheep:
my favourite idea is (if possible) to have a namespace that always resides between package namespaces and `.GlobalEnv`, and fill it with proxy functions that just call the `base` functions we want to shadow. while the kernel is running, we replace them.

janschulz:
I think the "env between packages and globalenv" is what RStudio does (AGPL...): 

https://github.com/rstudio/rstudio/blob/374557bdaf4423c016f41b6a43ffb2e9854fc862/src/cpp/r/R/Tools.R#L16

Exampleusage: : https://github.com/rstudio/rstudio/blob/374557bdaf4423c016f41b6a43ffb2e9854fc862/src/cpp/r/R/Options.R#L17

PR #324, opened 2016-05-11
Add special env to take r shadowing functions

janschulz:
Closes: #301

### DIFF
--- a/R/environment.r
+++ b/R/environment.r
@@ -1,0 +1,18 @@
+# Everthing related to the environment which takes functions which shadow base R functions.
+# This is needed to build in our own needs, like properly shutting down the kernel
+# when `quit()` is called.
+
+# The real env is created and attach'ed in the main() via the call to init_shadowenv()!
+init_shadowenv <- function(){
+    .irk.shadowenv <- attach(NULL, name = "jupyter:irkernel")
+
+    add_to_user_searchpath <- function(name, FN, attrs = list()) {
+        assign(name, FN, .irk.shadowenv)
+    }
+
+    # add the accessors to the shadow env to that env, so they are actually accessable 
+    # from everywhere...
+    add_to_user_searchpath(".irk.get_shadowenv", function() {.irk.shadowenv})
+    add_to_user_searchpath(".irk.add_to_user_searchpath", add_to_user_searchpath)
+    
+}
--- a/R/environment.r
+++ b/R/environment.r
@@ -2,17 +2,19 @@
 # This is needed to build in our own needs, like properly shutting down the kernel
 # when `quit()` is called.
 
-# The real env is created and attach'ed in the main() via the call to init_shadowenv()!
-init_shadowenv <- function(){
-    .irk.shadowenv <- attach(NULL, name = "jupyter:irkernel")
+add_to_user_searchpath <- function(name, FN) {
+    assign(name, FN, 'jupyter:irkernel')
+}
 
-    add_to_user_searchpath <- function(name, FN, attrs = list()) {
-        assign(name, FN, .irk.shadowenv)
-    }
+get_shadowenv <- function() {
+    as.environment('jupyter:irkernel')
+}
 
-    # add the accessors to the shadow env to that env, so they are actually accessable 
+# Adds functions which do not need any access to the executer into the users searchpath
+init_shadowenv <- function() {
+    # add the accessors to the shadow env itself, so they are actually accessable 
     # from everywhere...
-    add_to_user_searchpath(".irk.get_shadowenv", function() {.irk.shadowenv})
-    add_to_user_searchpath(".irk.add_to_user_searchpath", add_to_user_searchpath)
-    
+    add_to_user_searchpath('.irk.get_shadowenv', get_shadowenv)
+    add_to_user_searchpath('.irk.add_to_user_searchpath', add_to_user_searchpath)
+
 }
--- a/R/execution.r
+++ b/R/execution.r
@@ -115,7 +115,7 @@ quit = function(save = 'default', status = 0, runLast = TRUE) {
         if (!is.null(.GlobalEnv$.Last.sys)) .GlobalEnv$.Last.sys()
     }
     if (save) NULL  # TODO: actually save history
-    payload <<- c(payload, list(list(source = 'ask_exit', keepkernel = FALSE)))
+    payload <<- c(.self$payload, list(list(source = 'ask_exit', keepkernel = FALSE)))
 },
 
 handle_error = function(e) {
@@ -215,9 +215,9 @@ execute = function(request) {
     err <<- list()
     
     # shade base::quit
-    assign('quit', .self$quit, envir = .GlobalEnv)
-    assign('q',    .self$quit, envir = .GlobalEnv)
-    
+    .irk.add_to_user_searchpath('quit', .self$quit)
+    .irk.add_to_user_searchpath('q', .self$quit)
+
     # find out stack depth in notebook cell
     # TODO: maybe replace with a single call on first execute and rest reuse the value?
     tryCatch(evaluate(
@@ -250,12 +250,12 @@ execute = function(request) {
     # Workaround to warn user when code contains potential problematic code
     # https://github.com/IRkernel/repr/issues/28#issuecomment-208810772
     # See https://github.com/hadley/evaluate/issues/66
-    if(.Platform$OS.type == 'windows') {
+    if (.Platform$OS.type == 'windows') {
         # strip whitespace, because trailing newlines would trip the test...
         code <- gsub('^\\s+|\\s+$', '', request$content$code)
         real_len <- nchar(code)
         r_len <- nchar(paste(capture.output(cat(code)), collapse = '\n'))
-        if (real_len != r_len){
+        if (real_len != r_len) {
             msg = c('Your code contains a unicode char which cannot be displayed in your ',
                     'current locale and R will silently convert it to an escaped form when the ',
                     'R kernel executes this code. This can lead to subtle errors if you use ',
@@ -264,7 +264,6 @@ execute = function(request) {
             send_error_msg(paste(msg, collapse = '\n'))
         }
     }
-
     tryCatch(
         evaluate(
             request$content$code,
--- a/R/execution.r
+++ b/R/execution.r
@@ -215,8 +215,8 @@ execute = function(request) {
     err <<- list()
     
     # shade base::quit
-    .irk.add_to_user_searchpath('quit', .self$quit)
-    .irk.add_to_user_searchpath('q', .self$quit)
+    add_to_user_searchpath('quit', .self$quit)
+    add_to_user_searchpath('q', .self$quit)
 
     # find out stack depth in notebook cell
     # TODO: maybe replace with a single call on first execute and rest reuse the value?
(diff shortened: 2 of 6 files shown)


janschulz:
```

And that's how it looks like:

c:\data\external\R\IRkernel (envs)
λ jupyter console --kernel ir33
Jupyter Console 4.1.1


In [1]: search()
 [1] ".GlobalEnv"        "jupyter:irkernel"  "package:stats"
 [4] "package:graphics"  "package:grDevices" "package:utils"
 [7] "package:datasets"  "package:methods"   "Autoloads"
[10] "package:base"
```

R/kernel.r
```diff
@@ -381,6 +381,18 @@ main <- function(connection_file = '') {
...
     log_debug('Starting the R kernel...')
+    # Environment which will take functions which shadow main R stuff like q()
+    # This has to be done here, because a) it didn't work when I tried to add
+    # it as a notmal package item and not here in the codepath of the startup
+    # function and b) so the on.exit can detach the env when main is finished
+    .irk.env <- attach(NULL, name = "jupyter:irkernel")
+    on.exit({
+        detach("jupyter:irkernel")
+    })
+    assign(".irk.get_userenv", function() {.irk.env})
```
flying-sheep: where does this one end up?

janschulz: In .irk.env

janschulz: Or it should, will fix...

janschulz:
Latest code:

```
c:\data\external\R
λ jupyter console --kernel ir33
Jupyter Console 4.1.1


In [1]: print("x")
[1] "x"

In [2]: .ir
.irk.add_to_user_searchpath .irk.get_userenv

In [2]: .irk.get_userenv()
<environment: 0x000000000747cc78>
attr(,"name")
[1] "jupyter:irkernel"

In [3]: names(.irk.get_userenv())
[1] "quit"                        ".irk.get_userenv"
[3] ".irk.add_to_user_searchpath" "q"

In [4]: search
function ()
.Internal(search())
<bytecode: 0x0000000006d5d2f8>
<environment: namespace:base>

In [5]: search()
 [1] ".GlobalEnv"        "jupyter:irkernel"  "package:stats"
 [4] "package:graphics"  "package:grDevices" "package:utils"
 [7] "package:datasets"  "package:methods"   "Autoloads"
[10] "package:base"

In [6]: q()
Shutting down kernel

c:\data\external\R
λ
```

janschulz:
I think I have now created and reverted `environment.r` three times and each time I couldn't get it to work...

R/kernel.r
flying-sheep: haha

flying-sheep:
OK, what didn’t work?

* `on.exit` has of course to be inside of `main` because it triggers when the enclosing function returns
* did you try putting everything except `on.exit` into a function and calling that one in `main`?
* did you try getting the env by name every time you use it? i.e. create `env.name <- 'jupyter:irkernel'` and then do `assign('name', thing, env.name)`? then you can maybe do it in a separate file.

after all, it would be best to have all the stuff in `environment.r` as you tried, and the `assign` calls in `main`

R/kernel.r
```diff
@@ -381,6 +381,20 @@ main <- function(connection_file = '') {
...
+    # This has to be created here, because a) it didn't work when I tried to add
+    # it as a normal package item and not here in the codepath of the startup
+    # function and b) so the on.exit can detach the env when main is finished and
+    # c) it can't be created elsewhere because attach copies the env, so it would
+    # need to be reassigned afterwards...
+    .irk.env <- attach(NULL, name = "jupyter:irkernel")
+    on.exit({
+        detach("jupyter:irkernel")
+    })
+    assign(".irk.get_userenv", function() {.irk.env}, .irk.env)
```
takluyver: Naming quibble: 'user environment' is a lot like 'user namespace', which in IPython means the namespace where user code executes (i.e. typing `a = 1` defines a in the user namespace). That's not what we mean here. Can we come up with a clearer name? E.g. `get_shadowing_env`?

R/kernel.r
```diff
@@ -381,6 +381,20 @@ main <- function(connection_file = '') {
         connection_file <- commandArgs(TRUE)[[1]]
     }
     log_debug('Starting the R kernel...')
+    # Environment which will take functions which shadow main R stuff like q()
+    # This has to be created here, because a) it didn't work when I tried to add
+    # it as a normal package item and not here in the codepath of the startup
+    # function and b) so the on.exit can detach the env when main is finished and
+    # c) it can't be created elsewhere because attach copies the env, so it would
+    # need to be reassigned afterwards...
```
takluyver: I don't fully follow all this, but putting it in the `main` function feels like the wrong place. Can it be set up in the constructor of the Executor class?

R/kernel.r
janschulz: done

R/kernel.r
janschulz: I think no, because it has to use `on.exit` and the only functions which survives the whole lifetime of the kernel is `main()`

janschulz:
> OK, what didn’t work?

* implementing it as a "package environment" much like the one for the com manager and `on.exit` also on that level (though of it as "when the package is unloaded"). But putting it there, the code was never executing (surrounded it with prints) and `search()` never showed anything
* Creating the env manually via `new.env()` and only attaching it there, but having the rest of the functions defined elsewhere -> attach copies the env, so the functions in that place still had a reference to the old env, but that env was actually not in the search path but the copy... Was funny because `.irk.get_userenv()` was avaiölable (was copied into the new env) and returned the old env...

What could work is 
* define the `.irk.shadowenv` variable in `environment.r` and set initially to `NULL`
* define the functions as normal functions (-> can then only be used in the kernel package)
* add a `init_shadowenv()` functions which adds these functions to the shadow env
* then add `.irk.shadowenv <<- attach(null, name='jupyter:irkernel') (+ `on.exit`) in `main()`.

Will try that now...

janschulz:
Ah, and that doesn't work because of 

```
Error in init_shadowenv() :
  cannot change value of locked binding for '.irk.shadowenv'
Calls: <Anonymous> -> init_shadowenv
```

If someone wants to have a try, this is the best I can give you... :-/

flying-sheep:
you missed my point about using the env name and not the env object. that will probably work

janschulz:
Ok, I found a way to put it all into a function and call it from main... It's also not as nice, but at least all the code is in one place...

janschulz:
So, have a look and merge, I think this is ready...

takluyver:
Seems reasonable.

R/environment.r
```diff
@@ -0,0 +1,18 @@
+# Everthing related to the environment which takes functions which shadow base R functions.
+# This is needed to build in our own needs, like properly shutting down the kernel
+# when `quit()` is called.
+
+# The real env is created and attach'ed in the main() via the call to init_shadowenv()!
+init_shadowenv <- function(){
+    .irk.shadowenv <- attach(NULL, name = "jupyter:irkernel")
```
flying-sheep: '

R/environment.r
```diff
@@ -0,0 +1,18 @@
...
+
+    add_to_user_searchpath <- function(name, FN, attrs = list()) {
+        assign(name, FN, .irk.shadowenv)
+    }
+
+    # add the accessors to the shadow env to that env, so they are actually accessable 
+    # from everywhere...
+    add_to_user_searchpath(".irk.get_shadowenv", function() {.irk.shadowenv})
```
flying-sheep: `function() {.irk.shadowenv}` → `function() .irk.shadowenv`

R/kernel.r
```diff
@@ -381,6 +381,13 @@ main <- function(connection_file = '') {
         connection_file <- commandArgs(TRUE)[[1]]
     }
     log_debug('Starting the R kernel...')
+    # Create the shadow env here
+    # This has to be here so the on.exit can detach the env when main is finished
+    # = available for the whole lifetime of the kernel.
+    init_shadowenv()
+    on.exit({
+        detach("jupyter:irkernel")
```
flying-sheep: '

R/execution.r
```diff
@@ -215,9 +215,9 @@ execute = function(request) {
     err <<- list()
     
     # shade base::quit
-    assign('quit', .self$quit, envir = .GlobalEnv)
-    assign('q',    .self$quit, envir = .GlobalEnv)
-    
+    .irk.add_to_user_searchpath('quit', .self$quit)
```
flying-sheep: move to main?

flying-sheep:
jup, looks good! some questions and style things remain.

R/environment.r
```diff
@@ -0,0 +1,18 @@
+# Everthing related to the environment which takes functions which shadow base R functions.
+# This is needed to build in our own needs, like properly shutting down the kernel
+# when `quit()` is called.
+
+# The real env is created and attach'ed in the main() via the call to init_shadowenv()!
+init_shadowenv <- function(){
+    .irk.shadowenv <- attach(NULL, name = "jupyter:irkernel")
+
+    add_to_user_searchpath <- function(name, FN, attrs = list()) {
+        assign(name, FN, .irk.shadowenv)
```
flying-sheep: can you explain why don’t you use the environment name as i suggested?

flying-sheep:
jup, looks good! some questions and style things remain.

R/environment.r
janschulz: I should really invest some time into setting up some precommit hook...

flying-sheep: i tried to create some lintr rules, but somehow can’t get it to run: jimhester/lintr#139

R/execution.r
janschulz: I think there are a lot of things which are static between code executions, one could argue that these belong to the init method of the executer. -> later PR to cleanup that?

R/environment.r
janschulz: Probably because I don't know suche a method :-)

You eman there is something like `get_env_by_name('jupyter:irkernel')`?

R/environment.r
janschulz: That's another "habit" (apart from the smartquotes... :-)) of mine which I don't want to break: all codeblocks belong into brackets so that in the next "enter", the functions still works and does not need to be edited to get such brackets.

-> If there is error to have these brackets here, I would like to keep them...

R/environment.r
flying-sheep: no: `assign(name, FN, 'jupyter:irkernel')`

R/execution.r
flying-sheep: sure, why not! :smile:

R/environment.r
flying-sheep: it’s just R style. functions with one statement have no brackets. e do this everywhere, in every project, and others do as well.

it’s OK not to do it in cases like line 9 (assigning a function to a name), but when using a function as parameter, those braces just serve to make noise.

janschulz: >  but when using a function as parameter, those braces just serve to make noise.

We really have a different opinion on what looks good and what not :-) It really pains me to remove that...

R/environment.r
janschulz: because I didn't know that this was possible :-)

Seeing this, I think it is even possible to not init anything but just place these functions in the package...

R/environment.r
flying-sheep: well, as said: it’s like that everywhere in the R world. but if you’re OK with me some day forgetting and removing the braces, then leave it in :smile:

R/environment.r
flying-sheep: good idea!

janschulz:
Refactored again: now the thing is out of main() and in the executor. Also removed the space/tab problem.

R/environment.r
janschulz: I solved by making a normal functions form it and there I added braces...

BTW: gcc just got warnings to check for this (pseudo code):

```
if (whatever)
   do_something() # run only when whatever is true
   do_always() # run always
```

This would not have happend / needed if everybody would just always use braces :-)

takluyver: Yeah, my understanding is that - at least in C - omitting the parentheses is considered slightly bad practise.

flying-sheep:
perfect!

Issue #326, opened 2016-05-12
No rzmq binary for R 3.3.* or 3.4.*


richierocks:
When following the installation instructions, [here](https://irkernel.github.io/installation/) under the current R release, 3.3.0 or R-devel (which has version 3.4.0), no `rzmq` binary is found.

```{r}
install.packages('rzmq', repos = 'http://irkernel.github.io')
## Warning in install.packages :
##   cannot open URL 'http://irkernel.github.io/bin/windows/contrib/3.3/PACKAGES.gz': HTTP status was '404 Not Found'
```

Installing from source fails for me, but installing the binary under R-3.2.5 works.

Related: https://github.com/IRkernel/IRkernel/issues/298

janschulz:
Please don't use that anymore, just use the devtools version as in #327 until we have everything on CRAN... Sorry for the inconvenience...

PR #327, opened 2016-05-12
Change install instructions to devtools

janschulz:
The old instructions were still pointing to the rzmq package which we do not use
anymore and we didn't update our own repo with the latest versions.

This should "fix" issues like https://github.com/IRkernel/IRkernel/issues/326 by not pointing to our old repo...

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,14 @@ For detailed requirements and install instructions see [irkernel.github.io](http
 
 ## Installation
 
-We provide Windows and Mac OS X binary packages of all the needed packages:
+We are currently in the process of submitting our packages to CRAN. Until then, you can install via the `devtools` package:
 
-```r
-install.packages(c('repr', 'IRkernel', 'IRdisplay'),
-                 repos = c('http://irkernel.github.io/', getOption('repos')))
-IRkernel::installspec()
+```R
+install.packages(c('pbdZMQ', 'repr', 'devtools')) # repr is already on CRAN
+# devtools::install_github('irkernel/repr') # or get the latest repr from master
+devtools::install_github('irkernel/IRdisplay')
+devtools::install_github('irkernel/IRkernel')
+IRkernel::installspec() # to register the kernel in the current R installation
 ```
 
 Per default `IRkernel::installspec()` will install a kernel with the name “ir” and a
@@ -27,10 +29,10 @@ by supplying a `name` and `displayname` argument to the `installspec()` call (Yo
 install these packages in all interpreters you want to run as a jupyter kernel!):
 
 ```r
+# in R 3.3
+IRkernel::installspec(name = 'ir33', displayname = 'R 3.3')
 # in R 3.2
 IRkernel::installspec(name = 'ir32', displayname = 'R 3.2')
-# in R 3.1
-IRkernel::installspec(name = 'ir31', displayname = 'R 3.1')
 ```
 
 Now both R versions are available as an R kernel in the notebook.
@@ -48,7 +50,7 @@ If you have Jupyter installed, you can create a notebook using IRkernel from the
 You can also start other interfaces with an R kernel:
 
 ```bash
-# “ir” is the kernel name installed by the above 'IRkernel::installspec()'
+# 'ir' is the kernel name installed by the above 'IRkernel::installspec()'
 # change if you used a different name!
 jupyter qtconsole --kernel=ir
 jupyter console --kernel=ir


README.md
```diff
@@ -48,7 +50,7 @@ If you have Jupyter installed, you can create a notebook using IRkernel from the
 You can also start other interfaces with an R kernel:
 
 ```bash
-# “ir” is the kernel name installed by the above 'IRkernel::installspec()'
```
janschulz: No unicode in code...

README.md
```diff
@@ -12,12 +12,14 @@ For detailed requirements and install instructions see [irkernel.github.io](http
 
 ## Installation
 
-We provide Windows and Mac OS X binary packages of all the needed packages:
-
-```r
-install.packages(c('repr', 'IRkernel', 'IRdisplay'),
-                 repos = c('http://irkernel.github.io/', getOption('repos')))
-IRkernel::installspec()
+We are currently in the process to submit our packages to CRAN. Until then, you can install via devtools:
```
takluyver: 'in the process of submitting'

takluyver:
Other than that thing with the wording, looks fine to me.

janschulz:
Updated

takluyver:
OK, we'll give @flying-sheep a couple of hours to see this and IRkernel/irkernel.github.io#29, but I'm happy.

README.md
flying-sheep: oh, come ON!

janschulz: Sorry, I've so many C&P problems due to non-ascii chars in code examples, I'm firmly in the camp of not writing any unicode chars if I don't have to...

Yes, I'm on windows...

flying-sheep: do i understand correctly that pasting a *comment* with non-ASCII characters into a command line or notebook can cause issues of some type?

if so: WTF

if not: pls revert hunk

takluyver: I think that's probably fair enough with code samples. This is a bit of a corner case because it's a comment, but someone may still copy and paste the whole block, and I don't think smart quotes are that important. If it was someone's name with a non-ascii character in, I'd take it more seriously, but let's not waste time worrying about quotes.

flying-sheep: it’s a comment.

if a comment causes problems i want to know this or future reference.

if it doesn’t this change makes no sense and can be reverted.

takluyver: I don't know about R's unicode support, but in Python 2, using a non-ascii character in a source file without the encoding magic comment is a syntax error. In Python 3, it works so long as the file is saved in UTF-8, which may not be the default in all editors, especially on Windows. So there's clearly the potential for non-ascii characters to break stuff.

janschulz: I don't think it was a comment, but usually stupid webpages which had some JS part which replaced all quotes with smartquotes. But since then I associate smartquotes in code (including comments) with errors and more work and it feels kind of "dirty" (instead of your "beauty") and unprefoessional seeing them in source code parts. As if someone had programmed in MS word...

Beauty is in the eye of the beholder....

janschulz: BTW:

```R
>>> "# “ir” is"
"# ir is"
```
[In the notebook]

flying-sheep: “smart” means inserting them or is used as contrast to “dumb” (typewriter) quotes.

in fact they’re *real* quotes and typewriter quotes were originally a hack because typewriters were technically limited.

today, [the typographic standard is to use curly quotes for quoted phrases in text](http://practicaltypography.com/straight-and-curly-quotes.html) (citations, …). as a convention, backticks often get used for code, inside and out of markdown. and finally, typewriter quotes mean *string*. i.e. a piece of textual data.

so i hope you can shed your conditioning that the mistakes of bad tools imprinted on you, and come to appreciate good typography in time.

> BTW:
> 
> ```R
> >>> "# “ir” is"
> "# ir is"
> ```
> 
> [In the notebook]

sure, because you pasted it as string. string encoding in R on windows is fucked up, i know. if you’d have pasted it as comment, nothing would have happened because comments don’t get interpreted.

janschulz: And another FU stuff:

```
λ git slog
635a865 Add packageIQR repr
[...]
a27c12e don<E2><80><99>t double escape latex list names
```

flying-sheep: what terminal emulator do you use? cmd, cmder or powershell? what git distribution? what locale?

one of them is buggy and doesn’t support unicode, as the commit message is obviously fine: a27c12e51b24e87f91127bd11f668934bb211b7a

janschulz: I use cmder and `chcp 65001`. In this case it seems to be `less` which is to blame:
```
c:\data\external\R\repr (packageIQR)
λ git slog
2b62d01 Add packageIQR repr
[...]
a27c12e don<E2><80><99>t double escape latex list names

c:\data\external\R\repr (packageIQR)
λ git slog > test.txt # all is good and UTF8

c:\data\external\R\repr (packageIQR)
λ git slog |grep a27c12e
a27c12e don’t double escape latex list names

c:\data\external\R\repr (packageIQR)
λ git slog |grep a27c12e |less
a27c12e don<E2><80><99>t double escape latex list names
(END)

c:\data\external\R\repr (packageIQR)
λ git slog |grep a27c12e |more
Nicht genügend Arbeitsspeicher.
grep: write error
```

The last is the worst, because windows `more` refuses to run when `chcp 65001`...

flying-sheep: well, it’s 2016, unicode is 12 years old and projects are still using cp-something encodings instead of UTF-8. that’s just a wildly anachronistic environment you’re dealing with.

PR #330, opened 2016-05-13
WIP, don't review: More tests

janschulz:


flying-sheep:
this still contains the commit from #324…

maybe you need to rebase this on master?

janschulz:
I'm currently just trying if/how that works (currently it doesn't...)...

.travis-build.sh
```diff
@@ -35,7 +36,11 @@ travis_fold end Installing
 travis_fold start Python_env
  bootstrap_header 'Installing Python Environment for testing'
  # http://conda.pydata.org/docs/travis.html
- wget https://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh;
+ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then
+    wget https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh -O miniconda.sh;
+ else
+    wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh;
+ fi
```
flying-sheep: maybe:

```bash
[[ "$TRAVIS_OS_NAME" == "osx" ]] && os=MacOSX || os=Linux
wget "https://repo.continuum.io/miniconda/Miniconda3-latest-$os-x86_64.sh" -O miniconda.sh
```

janschulz: nice

janschulz:
So I think we run her into the same problems as all Mac OSx users:

```
0.00s$ cat "$INSTALL_LOG"
* installing *source* package ‘IRkernel’ ...
** R
** inst
** preparing package for lazy loading
Error : .onLoad failed in loadNamespace() for 'pbdZMQ', details:
  call: dyn.load(file, DLLpath = DLLpath, ...)
  error: unable to load shared object '/Users/travis/R/Library/pbdZMQ/libs/pbdZMQ.so':
  dlopen(/Users/travis/R/Library/pbdZMQ/libs/pbdZMQ.so, 6): Library not loaded: /Builds/CRAN-QA-Simon/packages/mavericks-x86_64/Rlib/3.3/pbdZMQ/libs/libzmq.4.dylib
  Referenced from: /Users/travis/R/Library/pbdZMQ/libs/pbdZMQ.so
  Reason: image not found
ERROR: lazy loading failed for package ‘IRkernel’
* removing ‘/Users/travis/build/IRkernel/IRkernel/IRkernel.Rcheck/IRkernel’
```

https://travis-ci.org/IRkernel/IRkernel/jobs/130009741

Do you think it makes sense to ask the pbdZMQ package to statically link to the underlying lib?

janschulz:
So, and then there is wondows: https://github.com/craigcitro/r-travis

flying-sheep:
eh?

> # NATIVE R SUPPORT IN TRAVIS IS LIVE!

janschulz:
Damn, wrong Link, have to find the right one again...

.travis-build.sh
```diff
@@ -7,6 +7,15 @@ bootstrap_header() {
   echo -e "${ANSI_YELLOW}$1${ANSI_RESET}"
 }
 
+travis_fold() {
```
flying-sheep: this already exists on travis. please remove it.

.travis.yml
```diff
@@ -13,15 +17,23 @@ env:
...
 r_github_packages:
   # we currently still need IRdisplay from github in all cases...
   - IRkernel/IRdisplay
 
 
 script:
-  - source .travis-build.sh
+  - export CHECK_LOG=${TRAVIS_BUILD_DIR}/IRkernel.Rcheck/00check.log
+  - export INSTALL_LOG=${TRAVIS_BUILD_DIR}/IRkernel.Rcheck/00install.out
+  - bash .travis-build.sh
```
flying-sheep: maybe the sourcing was necessary to access the travis_fold function. if so, please revert this line

.travis.yml
```diff
@@ -13,15 +17,23 @@ env:
...
   - grep -q -R 'WARNING' "$CHECK_LOG" ; [ $? -ne 0 ]
   # TODO: currently we do have 1 NOTE: Found the following assignments to the global environment...
   #- grep -q -R 'NOTE' "$CHECK_LOG"; [ $? -ne 0 ]
 
+after_failure:
+  - printf "travis_fold:start:show.log\n"
```
flying-sheep: 1. use the travis_fold fucntion instead of manually printing stuff
2. this doesn’t work properly. it’s run as separate commands and messes up stuff. you need to extract this into a shell script.
3. are you sure those logs aren’t already in some folded section?

.travis-build.sh
janschulz: yep, that was a problem when I executed instead of sourced the file

flying-sheep: OK, then please remove the function and go back to sourcing it.

janschulz: AFAIR the reason for not sourcing was a problem with `set -e` which terminated the whole build including the `after_failure` clause in the travis script. Running it as `bash script` and setting set -e fixed that.

Without `set -e` you have to check each command so that the build fails if one of the steps fails (e.g. the problem you noticed that the test builds do not show that something is wrong...).

I will try to get a fix for both problems... Going back to a "travis.yml only" setup would be my preference... IMO one big travis.yml is better than a travis.yml which links to a big script. Especially if the amount of code is reduced because we don't need to reimplement most of travis' r support?

janschulz: See also here: https://docs.travis-ci.com/user/customizing-the-build/#Implementing-Complex-Build-Steps -> calling the script with `set -ev`

flying-sheep: > Going back to a "travis.yml only" setup would be my preference...

i tried it, but i didn’t find a way to get folding then. separate shell lines in YAML are treated as separate commands, which messes with the folding, …

maybe we can put it all in the YAML like this?

```yaml
script:
  - |
    travis_fold start Building
    bootstrap_header 'Building package'
    R CMD build .
    travis_fold end Building
  - ...
```

janschulz: That looks like a great idea... Will try that, but it could be a few days until I can come back to this... I also would like to have a solution for the zmw problem on OSX first...

janschulz:
This is usefull: https://lint.travis-ci.org/ ...

flying-sheep:
you think so?

> in language section: illegal value r, defaulting to ruby

this is obviously wrong. travis has first-class R support.

janschulz:
R is not officially supported... I had more problems which ended up not even starting a travis run... for that it was great :-)

flying-sheep:
ah OK. btw.: don’t bother potting it all back into `.travis.yml` if you don’t manage to get the folding to work. better check this first before you invest too much time.

janschulz:
Folding is coming back...

flying-sheep:
great! will it also fail properly or does only the return value of the last command (which would be the folding) count?

janschulz:
It looks like:

* Having it in a big list (using  `- | ...`) or in sourced form or with using a helper file which defines travis helper functions **without** `set -x`, we need to check each ran command and call something like travis_terminate (no idea if that is defined)
* `set -x` will terminate the complete travis job without running the `after_failure` clause  (=no logfiles are cat'ed :-()
* using normal one command per line won't get us folding...

I'm currently leaning to use the last solution: using one command per line and putting all the setup code into the `before_install`, so this (conda setup, r packages) is at least "folded away" and in the normal install only the checks are done.

janschulz:
Ok, the main commits are done, now I only want to get a test for the plots in...

janschulz:
Ok, no test for the Mac OS X death:

```py

class KillsOSXTests(AbstractIRKernel):
    def test_kills_macosx(self):
        "See IRkernel/IRkernel#342"
        code = """
install.packages("ggplot2")
library(ggplot2)
qplot(1:10)
plot(1:10)
"""
        reply, output_msgs = self._execute_code(code)
        print(output_msgs)
        print(reply)
        self.assertEqual(len(output_msgs[0]), 2)
        
```

Can't get it to run right :-(

janschulz:
And it looks like it worked :-)

janschulz:
JFYI: this PR contains the revert for the "/dev/null png device" commit...

flying-sheep:
> JFYI: this PR contains the revert for the "/dev/null png device" commit...

what? why?

flying-sheep:
and the travis log is worse now. dozens of cryptic unwrapped commands.

before, we had a few neatly folded sections.

please make it look like before.

janschulz:
Revert: two reasons, it breaks Mac Users (See other issues) and it is noch `R CMD check` clean. :-)

For the better Output: See above, basically we have to die one death (is that actually a valid English saying?)

flying-sheep:
not so fast. it’s clean now (95bd51a3ea5bba17900a14138e02b3f9e73cbe65), and `/dev/null` is POSIX specified, so i need to see a *really good* argument why it should fail.

let’s phrase our goals:

1. stop failing immediately if there’s an error. we want to see the whole thing
2. we don’t want to check every return value
3. don’t change the output

the status quo is that we don’t do 1.; this is bad for making changes since we can’t see multiple fails at once.

we could have easily changed this to not do 2.; that would be bad beause we *will* forget to do it when adding more stuff.

your PR’s state is that you don’t do #3, which is bad to see quickly where a fail happens and what the basic things that are tested are.

---

so the only alternative i see is

1. create a `.travis` directory
2. put every section as .sh script in there. `set -e` in each one
4. ```yaml
    script:
      - export -f travis_fold # do not repeat ourselves
      - .travis/github_dependencies.sh
      - .travis/build_package.sh
      - .travis/check_package.sh
      ...
    ```

takluyver:
Regarding `/dev/null`, I agree that we should investigate more before changing anything, but the fact that we got a bug report almost as soon as we introduced that code makes me worry that it's going to be problematic. So I think unless we can find a clear reason that that was a weird outlier, we should err on the side of avoiding it, as annoying as that is.

flying-sheep:
that’s a reasonable line of thought. on the other hand i haven’t found anything about any problems with `/dev/null` on OSX in the internet, all problems people have were due to fucking with the file. let’s first have the issue reporter say something about his system before doing anything. if the file has the right node type and permissions on his system (or the wrong one for all OSX users), then there’s probably a problem for us.

so this PR can be merged if the three goals in my previous comment are fullfilled, and if a) either the `/dev/null` thing can be confirmed to be a systematic OSX problem, or b) the reverting commit is removed.

janschulz:
I removed the revert after the fixes for `R CMD check` are in.

The current status is now:
* all tests under  `scripts` are run, but if any of the tests fail, the build is marked as failed and the `after_failure` clause is run
* the output isn't collapseable

Regarding the reorg: I really hate such a  `.travis` directory style :-/ I already dislike the current setup with the two install functions...

I can still make it a bit shorter by redirecting the output of build and check to /dev/null so that this is only shown at after-failure.

Is scrolling so bad?

flying-sheep:
yes. clarity of output is important. the whole idea is to quickly see why a build went wrong and to be able to reason about the individual steps.

elegance of file organization can be pursued if it’s possible without compromising on other things. a bonus goal if you will

janschulz:
ok. I thought some more about this but I currently do not find a way to do that without writing code which I find "suboptimal":

If individual steps should be folded in the log, then this needs to be done in a way that `fold start/stop` sequence is not broken. But that means you basically have to use a wrapper approach or collect the error term after each command so that the last `fold end` somehow returns the "unsuccessful" return code. This means that we would need to use the `command || ERROR=true` style of code. Or make `wrapper real-command` work (which probably involves eval?). Both something I don't want to see in the travis codepath, I already don't like the additional workarounds we take for the install stuff (defining them in an additional file just so that they are separated and the lines do not show up in the log)

simple, clean, working code >> clean log files

So IMO this is the best I'm able (and more or less willing) to create, unless you have some ideas how to make it work without compromising the simplicity (if someone can still say that to our setup) of the testing code.

Maybe the /dev/null redirect so that it becomes less scrolling...

flying-sheep:
did you try what i said (below the line in https://github.com/IRkernel/IRkernel/pull/330#issuecomment-220569867)?

didn’t it work?

janschulz:
You mean 

> put everything that can/should fail as a unit (every section) into a .sh script in there. set -e in each one

?
This would mean that the script starts with a `fold start` would get an error and return *without* calling the `fold end`. I've not investigated what travis makes of that, probably leave that part unfolded/open and maybe the next parts then as well.

flying-sheep:
ah, you’re of course right!

but i have a solution:

```bash
fold_wrap() {
	local header=$1; shift
	local section=$1; shift
	local cmd=$@
	travis_fold start $section
	headline $header
	"$cmd"
	local result=$?
	travis_fold end $section
	return $result
}

**Comment by janschulz** [comment]
Sorry, no.

IMO https://travis-ci.org/IRkernel/IRkernel/jobs/132126586 looks just fine and I don't see a benefit from puting in another level of abstraction just to get some shorter page. Scrolling down and then looking up for a red line seems perfectly fine.

The time I spent on working on this (I just tried to make `set -e` work in bash functions with your above wrap script, which seems not to work, so a travis subdir seems to be the last way) requirement is probably more than I will ever spent on scrolling down on any travis pages in any github project I work on.

So: no, I won't work anymore on this, I find it fine as it is.

re subdir + individual shell scripts: I really hate the idea of multiple twoline shells scripts in a confusing subdir where for every change I have to look into different scripts to find out where the specific test is now.

**Comment by janschulz** [comment]
I've also move the install steps back as my understanding of set -e and bash functions is currently, that they don't affect it: a error in one of the steps would not jump out of the function and therefore would not cancel the build.

test.sh:
```sh
fold_wrap() {
    local header=$1; shift
    local section=$1; shift
    local cmd=$@
    echo start $section $header
    "$cmd"
    local result=$?
    echo end $section
    return $result
}

example_command() {
	set -e
	echo Hallo
	false
	echo SHOULD NOT HAPPEN
	set +e
}
```

```bash
$ source test.sh
$ fold_wrap HEAD SEC example_command || echo "BAD"
start SEC HEAD
Hallo
SHOULD NOT HAPPEN
end SEC
```

flying-sheep:
the problem isn’t (only) scrolling, it’s the obscurity of some of the lines. a flding sections that says “testing R CMD check logs for items” before something like `grep -q 'Status..1 NOTE' "$CHECK_LOG"` is valuable.

but yeah: i veto the huge block of obscure commands with no wrapping and helpful info. i’m sure we’ll find a solution though.

how about we use `|| exit 1` in a function for the two-liners and bash scripts for the longer things?

janschulz:
As I said: I find it fine as is. I've searched through matplotlibs errors and that is a lot worse.

Ask yourself: how much time do you spend scrolling vs the time you spend adding that folding (and breaking the tests for several days/weeks until we noticed)... 

Simple travis.yml are best and we are already way past simple...

takluyver:
I'm inclined to side with simplicity here over the presentation of test runs. Folding sections of output is nice if it's convenient, but let's not spend too much time on it if that's the only obstacle. We can always come back to it later if Travis changes something or we think of a neater way to do it.

flying-sheep:
it’s easy, it just requires to split it up into files which @janschulz doesn’t like to do.

takluyver:
Split tests up, or split test infrastructure up? I also would rather not have a bunch of files just for running Travis.

flying-sheep:
me neither but it’s better than having the build terminate or a messy output.

test_ir.py
```diff
@@ -37,5 +37,68 @@ class DisplaySystem(jkt.KernelTests):
     {'code': '1:3', 'mime': {'text/plain':'[1] 1 2 3'}},
     ]
 
+class AbstractIRKernel(jkt.KernelTests):
```
takluyver: Doing it like this with a separate class for each test means that it re-discovers and then skips the base jupyter_kernel_test tests for each subclass. Is that intentional?

takluyver:
Let's merge this with the Travis infrastructure as @janschulz has done it, and then you can do a PR to make the output how you want, and we can weigh up the added complexity versus the messiness of the output.

flying-sheep:
actually i’m happy with the current tradeoff compared to the changed version.

so i’d say we leave the structure as it is (only merge the other changes now) or settle on a solution *before* we merge this, then only discuss details (but nothing fundamental) in my PR.

test_ir.py
janschulz: My intention was that the kernel was restarted for each test... as I play with options it seems to do that.

takluyver: On 23 May 2016 at 16:43, Jan Schulz <notifications@github.com> wrote:

> My intention was that the kernel was restarted for each test... as I play
> with options it seems to do that.
>
OK. By default JKT reuses a kernel between tests, because starting it up
and spinning it down can be slow, and the tests are written not too depend
on order. If you want to create a new kernel for each test, I'd recommend
using setUp and tearDown methods - essentially equivalent to the setUpClass
and tearDownClass methods that are already there:
https://github.com/jupyter/jupyter_kernel_test/blob/6c392ddf54a24c464fa5e07ad7fb001023f04698/jupyter_kernel_test/__init__.py#L21

janschulz:
>everything above 2-3 trivial lines goes into .travis/name_of_step.sh, in which set -e is used. the trivial steps go into functions in which the individual functions are guarded via || exit 1. scripts and functions are called using my wrapper above. are you both OK with this?

No: simple travis >> good-looking test logs

test_ir.py
janschulz: Ok, I reordered... somehow the kernel tester looks like it would be a good fit for pytest fixtures :-)

takluyver: I prefer to keep JKT at the simple common layer of unittest, but you've seen how little code it is to start a kernel, so if you want to make a pytest fixture for it, I'm fine with that. We could add it to JKT as a module.

flying-sheep:
“No” won’t do :laughing: 

do you have a better idea how to organize this? i’m seriously not willing to merge anything as underdocumented and unclear as this.

i have to spend a few seconds on each ine to see what it does and to mentally reconstruct the rationale why it is there. slapping on comments on each step won’t improve this much.

if you want this merged you’ll need to agree to some solution that structures the commands into a small number of steps that can be folded and have a good description.

takluyver:
The `before_install` section looks reasonable to me, but the `script` section is a bit opaque. I don't know how much it's possible to simplify that if we want to test the output from `R CMD check`, though.

flying-sheep:
[HA YES](http://stackoverflow.com/a/23806662/247482)!

```yaml
script:
  - |
    wrap_cmd() {
      ...
    }
  - |
    some_step()
      trap 'trap - ERR; return' ERR
      do_stuff
      idk_lol
    }
  - wrap_cmd 'Some step' some_step
```

takluyver:
Hmm, bash functions using trap statements inside a YAML file is a long way from what I'd call elegant ;-)

flying-sheep:
that’s only because bash isn’t elegant. and swallows errors by default.

i’d prefer just sourcing one big bash file, this way we avoid subtle YAML encoding gotchas (if such a thing exists) and gain syntax highlighting at the cost of having one more file.

what do you both think?

janschulz:
I want a single line per test in the travis.yaml and certainly not a wrapper. This has the disadvantage that the log will be slightly longer and more verbose (but it's still a long way from what I know in other projects), but it is a lot less prone to failure (=current situation). And it doesn't waste developer time when this has to be implemented (e.g. I think the time we both spent on this issue arguing is more that we both will ever spent on scrolling through the log).

IMO this PR is "ready to be merged" as it is.

I'm not going to implement a wrapping style, feel free to do it yourself, but also make sure that it fails when one of the steps go wrong, so that we don't end up in the current situation where we basically had no tests.

flying-sheep:
there’s also [an alternative which is more clear](http://stackoverflow.com/a/23808157/247482)!

```bash
some_step() (
	set -e
	im_in_a_subshell
)
```

flying-sheep:
hey, sorry to be confrontative, but you really need to read my stuff more carefully.

i said multiple times that i *only* want you to agree on me changing stuff up afterwards, not implement it yourself.

now you propose the same thing. you could simply have said from the beginning that you’re OK with whatever i do as long as i don’t create multiple files.

so if you’re both OK with the following course of action, please say so and we can go ahead:

1. merge this
2. i move everything into a separate .sh file to gain syntax highlighting (optional, please state your opinions)
3. i move all sections into individual functions (with `set -e` in a subshell so the whole function fails if a command fails)
4. i call those functions with a wrapper that adds folding inside of .travis.yml

it would look like:

```bash
#.travis.sh
wrap_cmd() { ... }
step_a() (
	set -e
	...
)
...
```

```yaml
#.travis.yml
script:
	- source .travis.sh
	- wrap_cmd 'Step A' step_a
	- ...
```

OK? @janschulz @takluyver?

takluyver:
That sounds fine to me. I mean, it sounds totally crazy that you can define functions with different kinds of brackets and they work in deeply different ways, but that's bash for you.

janschulz:
As I said: My vote goes to keeping it as is in this PR without putting in a wrapper and voodoo magic just for the sake of a shorter log and a few less seconds per log read. But I also don't want to waste any more time on this than I already have.

If you want to spend the time:

>  make sure that it fails when one of the steps go wrong, so that we don't end up in the current situation where we basically had no tests.

Please also document the different kind of brackets and so on in all places, that no one is accidentally breaking it in a later change to the test system. I didn't even know that such stuff exists.

takluyver:
Alright, thanks both.

flying-sheep:
in bash, `foo() compound_command` defines a function. [compound commands](https://www.gnu.org/software/bash/manual/bashref.html#Compound-Commands) can be loops, conditionals, and groupings.

the two kinds of groupings are those: `{}` is a command list (same environment/shell) and `()` is a subshell (new, inheriting environment/shell). you can use both to e.g. redirect collected output into a single command:

```bash
{
	echo '<!doctype html><html>'
	produce_html
	echo '</html>
} | expect_html_document`

i use subshells often interactively for this:

```bash
(cd somewhere; do_command)
```

which will not change my directory if the command fails.

bash is fucking complex. i also didn’t know that any kind of compound command is accepted as function body, but i think everyone should know about subshells.

Issue #267, opened 2016-02-25
Running vignette("nse") does not work


janschulz:
It opens a browser tab with `http://127.0.0.1:31903/library/dplyr/doc/nse.html`, but the tab doesn't load anything... :-(

flying-sheep:
uh, how is that related to IRkernel?

doesn’t `vignette` spin up its own webserver regardless of from where you call it?

JanSchulz:
Basically because it works in rstudio (shows up in rstudio in the area where the plots also show up) :-)

flying-sheep:
because RStudio intercepts many things (just like is)

what does vanilla R do for you though?

JanSchulz:
Also works...

flying-sheep:
so it spins up the server in both cases but only the plain R one works somehow?

wat

JanSchulz:
Vanilla R opens a browser which displays the thing. Here is tcpview of these two processes: the upper one is the one from a R console and the lower one from the R kernel in a notebook (which does not show a page).

![2016-02-28_214433](https://cloud.githubusercontent.com/assets/890156/13381771/8e5f10d2-de64-11e5-8720-c2dfc8e8e032.png)

JanSchulz:
Is this working on linux/mac?

JanSchulz:
Another interesting artefact is that if I restart the kernel (aka kill the old kernel), the vignette of the page which was opened by the **old** kernel is then displayed instantly. Even interrupting does work, so I have a nice workaround how to display a vignette: run the cell, wait for the new tab to open (which happens instantly but sits there and  "waits" for the content) and then `kernel -> interrupt`. Fun :-)

JanSchulz:
What is actually running the server?  I tried `vignette` to display the code but that seems to return a class  but does not spin up a server?

flying-sheep:
well, that would be `utils:::print.vignette`, of course!

since `repr` falls back to calling `print` when it has no special handling implemented, that explains why *something* happens. but not why it’s so weird.

we probably have to implement a better `repr_html.vignette`.

takluyver:
It sounds like the server it spins up is getting blocked by the kernel somehow - like when the kernel is inside the poll system call, that stops the server from responding.

I agree that the best way to solve it is probably to do something else to display vignettes.

JanSchulz:
If the kernel interferes with http servers, then we should test plotting libs which also spin up a server (if there are such libs)

takluyver:
[...]

Issue #280, opened 2016-03-16
View window not responding


renkun-ken:
Create a new R notebook and run

```r
View(mtcars)
```

Then the built-in data frame Viewer window shows but the window does not respond to user, I even cannot close the window. If I force closing the window, the R session will terminate too.

My session info:

```
R version 3.2.4 (2016-03-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 10586)

locale:
[1] LC_COLLATE=English_United States.1252 
[2] LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252
[4] LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] magrittr_1.5    IRdisplay_0.3   tools_3.2.4     base64enc_0.1-3
 [5] uuid_0.1-2      stringi_1.0-1   rzmq_0.7.7      IRkernel_0.5   
 [9] jsonlite_0.9.19 stringr_1.0.0   digest_0.6.9    repr_0.4       
[13] evaluate_0.8.3 
```

flying-sheep:
we should fix that by providing our own `View` and `edit` functions. good call!

renkun-ken:
That's great!

JanSchulz:
As a stop gap measure, we should replace these functions (and whatever does the input thing) with functions which raise an error.

ggrrll:
while waiting for `View()` , one can also use `head()` ...that is still quite useful ...

gvelasq:
Is there an active plan to implement these functions in `IRkernel`?

It could be helpful in `vscode-jupyter` as well as in the VS Code R extensions as discussed [here](https://github.com/microsoft/vscode-jupyter/issues/4604).

flying-sheep:
So is there an official message type in the protocol we can use for View?

payloads (including the `page` and `edit`) are deprecated and not supported by jupyter lab: https://jupyter-client.readthedocs.io/en/latest/messaging.html#payloads-deprecated

Of course I can make `View` just use `IRdisplay::display`, but I think `edit` isn’t possible without coding up an extension.

gvelasq:
Thanks for considering. I can't comment on the details of the implementation since it's beyond my expertise, but as a user I can say that the priority would be to implement `View()`.

[...]

Issue #85, opened 2015-04-15
Is there plan to make browser() and debug() work in Jupyter


changcw83:
currently, the kernel just get stuck when issuing the command.
Really looking for it as I really prefer using this to Rstudio. Just missing the browser and debug functions

JanSchulz:
http://www.stats.uwo.ca/faculty/murdoch/software/debuggingR/debug.shtml

JanSchulz:
Sounds like it is missing the "get input from the kernel client" feature is not yet implemented?

This is what the python kernel does:
* [getpass()](https://github.com/ipython/ipython_kernel/blob/master/ipython_kernel/kernelbase.py#L620)
* [raw_input()](https://github.com/ipython/ipython_kernel/blob/master/ipython_kernel/kernelbase.py#L637)
* [the actuall message sending](https://github.com/ipython/ipython_kernel/blob/master/ipython_kernel/kernelbase.py#L654)

Both methods are [put into the global builtin](https://github.com/ipython/ipython_kernel/blob/master/ipython_kernel/ipkernel.py#L133), so are available as `getpass()` and `raw_input()`

So what would be needed here is a way to replace the watever is used to read in text from stdin in R... No idea what/where this is...

katosh:
The R function `browser()` acts like a stopping point and enables debugging anywhere in the code. Currently jupyter would not stop when `browser()` is called but instead start printing out every line it evaluates. Is there a way to change the behavior to that of pythons pdb?

Using the R-magic in a python notebook with `%%R`, browser actually allows to send commands and the output is printed after exiting it with `Q`. This is already more helpful than the current behavior in my R notebooks.

ChrisRobo:
Try running:
```R
interactive()
```

I did it in a jupyter notebook with an R kernel just now and it returns `FALSE`.  R, frustratingly, doesn't respect `browser()` statements when it's in "non-interactive" mode.

flying-sheep:
There’s nothing we can do with the current architecture, `interactive()` means too many things at once. The way around it iirc is to make IRkernel a natively compiled package calling into C like Juniper.

janxkoci:
See also the new [Jupyter Debug Protocol](https://blog.jupyter.org/a-visual-debugger-for-jupyter-914e61716559) ;)

JZL:
[...]

PR #333, opened 2016-05-13
Add a few worakrounds for problematic r functions

janschulz:
* Workaround a problem with vignette(): closes: https://github.com/IRkernel/IRkernel/issues/267
* we have no `edit()`and `View()`: https://github.com/IRkernel/IRkernel/issues/280


R/environment.r
```diff
@@ -17,4 +17,26 @@ init_shadowenv <- function() {
     add_to_user_searchpath('.irk.get_shadowenv', get_shadowenv)
     add_to_user_searchpath('.irk.add_to_user_searchpath', add_to_user_searchpath)
 
+    # For the rest of the functions, please explain why the workaround is needed
+    # (=the problem) and link to the issue describing the problem.
+
+    # workaround for problems with vignette(xxx) not bringing up the vignette
+    # content in the browser: https://github.com/IRkernel/IRkernel/issues/267
+    add_to_user_searchpath('print.vignette', function(...){
+        utils:::print.vignette(...)
```
flying-sheep: ::: means private access. is this a private function?

janschulz:
```R
> utils::print.vignette
Error: 'print.vignette' is not an exported object from 'namespace:utils'
> 
```

takluyver: Is that going to be a problem for the checks CRAN does?

flying-sheep: yes, that’s why i mentioned it.

but we can use [this hack](https://github.com/IRkernel/repr/blob/df33b1574c8abb5eb7c19e7cc33fafbd60847e13/R/repr_help_files_with_topic.r#L12)

takluyver: Get your package on Cran with this one weird trick! ;-)

janschulz: done

janschulz:
Anything else?

janschulz:
https://github.com/IRkernel/IRkernel/issues/85
-> implement getting raw input from the frontend... But that's IMO for a different PR because it touches the messaging system.

So IMO this is ready for review and eventual merging...

R/environment.r
```diff
@@ -11,10 +11,36 @@ get_shadowenv <- function() {
...
     # from everywhere...
     add_to_user_searchpath('.irk.get_shadowenv', get_shadowenv)
     add_to_user_searchpath('.irk.add_to_user_searchpath', add_to_user_searchpath)
 
+    # For the rest of the functions, please explain why the workaround is needed
+    # (=the problem) and link to the issue describing the problem.
+
+    # workaround for problems with vignette(xxx) not bringing up the vignette
+    # content in the browser: https://github.com/IRkernel/IRkernel/issues/267
+    add_to_user_searchpath('print.vignette', function(...){
```
flying-sheep: `){` → `) {`

R/environment.r
```diff
@@ -11,10 +11,36 @@ get_shadowenv <- function() {
...
+        # returning immediately will run into trouble with zmq and its polling
+        # preventing the vignette server to startup. So wait a little to let
+        # it startup...
+        # 0.1 is too little, so add some margin...
+        Sys.sleep(0.5)
+    })
+
+    # we simply have currently no way to view or edit dfs:
+    # https://github.com/IRkernel/IRkernel/issues/280
+    add_to_user_searchpath('View', function(...){
```
flying-sheep: `){` → `) {`

R/environment.r
```diff
@@ -11,10 +11,36 @@ get_shadowenv <- function() {
...
+        stop("'View()' not yet supported in the Jupyter R kernel")
+    })
+    add_to_user_searchpath('edit', function(...){
```
flying-sheep: `){` → `) {`

flying-sheep:
looks great! have you confirmed that they works? then i’d just merge this…

R/environment.r
takluyver: :expressionless:

R/environment.r
janschulz: I switched to rstudio to lint the code and it doesn't show me this. Do we really want to waste time and effort on something which the RStudio inter doesn not catch?

janschulz: done BTW...

R/environment.r
janschulz: done

R/environment.r
janschulz: done

janschulz:
both work:

```
>>> View(data.frame())
Error in View(data.frame()): 'View()' not yet supported in the Jupyter R kernel
Traceback:

1. View(data.frame())
2. stop("'View()' not yet supported in the Jupyter R kernel")

>>> vignette("knitr")
starting httpd help server ... done
# and the vignette is displayed
```

Issue #306, opened 2016-04-25
set cran mirror for the session


janschulz:
Currently `install.packages` raises an error when called without explicitly setting a cran mirror. I would like to propose that we set one for the user, namely `https://cloud.r-project.org/`, which seems to redirect based on your current geo location: https://cran.r-project.org/mirrors.html

Setting should only happen if the users hasn't set a mirror in the profile file...

janschulz:
Here is our FAQ entry which would become obsolete: http://irkernel.github.io/faq/

flying-sheep:
how should we go about this?

check if `identical(getOption('repos'), c(CRAN = '@CRAN@'))` and set it then [here](https://github.com/IRkernel/IRkernel/blob/cb7fd617fda7959f0855623a2529d36bd7c93d45/R/execution.r#L308-L326)?

janschulz:
This is what RStudio does:

https://github.com/rstudio/rstudio/blob/f337647969e2c02394d4dac06212aa8501777025/src/cpp/r/R/Tools.R#L484-L520
```R
.rs.addFunction( "setCRANRepos", function(reposUrl)
{
  local({
      r <- getOption("repos");
      # attribute indicating the repos was set from rstudio prefs
      attr(r, "RStudio") <- TRUE
      r["CRAN"] <- reposUrl;
      options(repos=r)
    })
})

.rs.addFunction( "setCRANReposAtStartup", function(reposUrl)
{
   # check whether the user has already set a CRAN repository
   # in their .Rprofile
   repos = getOption("repos")
   cranMirrorConfigured <- !is.null(repos) && repos != "@CRAN@"

   if (!cranMirrorConfigured)
      .rs.setCRANRepos(reposUrl)
})


.rs.addFunction( "isCRANReposFromSettings", function()
{
   !is.null(attr(getOption("repos"), "RStudio"))
})


.rs.addFunction( "setCRANReposFromSettings", function(reposUrl)
{
   # only set the repository if the repository was set by us
   # in the first place (it wouldn't be if the user defined a
   # repository in .Rprofile or called setRepositories directly)
   if (.rs.isCRANReposFromSettings())
      .rs.setCRANRepos(reposUrl)
})
```

Lets just do something similar in a new `init_session()`?

janschulz:
PR incoming...

flying-sheep:
> `!is.null(repos) && repos != "@CRAN@"`

[...]

PR #339, opened 2016-05-19
Set default repo during the kernel session

janschulz:
We set repos to the default mirror of CRAN (which has geo-redirect enabled) aswe currently do not have any way to ask the user in case it is unset.
    
The setting is only set if it is unset (e.g. the user does not set it in .RProfile).

Closes: https://github.com/IRkernel/IRkernel/issues/306

### DIFF
--- a/R/environment.r
+++ b/R/environment.r
@@ -44,3 +44,20 @@ init_shadowenv <- function() {
         stop("'edit()' not yet supported in the Jupyter R kernel")
     })
 }
+
+init_cran_repo <- function() {
+    r <- getOption('repos')
+    is_unuseable_mirror <- identical(r, c(CRAN = '@CRAN@'))
+    if (is_unuseable_mirror) {
+        # the default repo according to https://cran.r-project.org/mirrors.html
+        # uses geo-redirects
+        r[['CRAN']] <- 'https://cran.r-project.org'
+        # attribute indicating the repos was set by us...
+        attr(r, 'irkernel') <- TRUE
+        options(repos = r)
+    }
+}
+
+init_session <- function() {
+    init_cran_repo()
+}
--- a/R/execution.r
+++ b/R/execution.r
@@ -320,7 +320,11 @@ initialize = function(...) {
     # Create the shadow env here and detach it finalize
     # so it's available for the whole lifetime of the kernel.
     attach(NULL, name = 'jupyter:irkernel')
+
+    # Add stuff to the user environment and configure a few options
+    # in the current session
     init_shadowenv()
+    init_session()
 
     callSuper(...)
 },
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -11,7 +11,7 @@ other frontends) submits to the kernel via the network.
 
 ## R CMD check results
 
-There are no WARNINGs and 1 NOTEs.
+There are no ERRORS and no WARNINGs and 1 NOTEs.
 
 Note:
 
@@ -23,7 +23,7 @@ Note:
   We use this additional environment to add functions so that regular "stuff"
   like `quit()` works in the IRkernel environment (in this case to shutdown the
   kernel). We added the "good practice" call to `detach` in `finalize` so that
-  this environment is available as long as the R kernel is running..
+  this environment is available as long as the R kernel is running.
 
 ## Downstream dependencies
 


cran-comments.md
```diff
@@ -11,7 +11,7 @@ other frontends) submits to the kernel via the network.
 
 ## R CMD check results
 
-There are no WARNINGs and 1 NOTEs.
+There are no ERROS and no WARNINGs and 1 NOTEs.
```
takluyver: errors

R/environment.r
```diff
@@ -44,3 +44,23 @@ init_shadowenv <- function() {
         stop("'edit()' not yet supported in the Jupyter R kernel")
     })
 }
+
+
+init_cran_repo <- function() {
+    local({
+        r <- getOption('repos');
+        is_configured_by_user <- !is.null(r) && r != '@CRAN@'
```
flying-sheep: use identical as i said, please.

R/environment.r
```diff
@@ -44,3 +44,23 @@ init_shadowenv <- function() {
...
+            # uses geo-redirects
+            r['CRAN'] <- 'https://cran.r-project.org'
+            # attribute indicating the repos was set by us...
+            attr(r, 'irkernel') <- TRUE
+            options(repos = r)
+        }
+    })
+}
+
+init_session <- function() {
```
takluyver: Is this name something special that R calls automatically?

R/environment.r
```diff
@@ -44,3 +44,23 @@ init_shadowenv <- function() {
         stop("'edit()' not yet supported in the Jupyter R kernel")
     })
 }
+
+
+init_cran_repo <- function() {
+    local({
+        r <- getOption('repos');
```
flying-sheep: aaargh a semicolon! :mask:

R/environment.r
```diff
@@ -44,3 +44,23 @@ init_shadowenv <- function() {
...
+        is_configured_by_user <- !is.null(r) && r != '@CRAN@'
+        if (!is_configured_by_user) {
+            # the default repo according to https://cran.r-project.org/mirrors.html
+            # uses geo-redirects
+            r['CRAN'] <- 'https://cran.r-project.org'
```
flying-sheep: double brackets to set a single element.

R/environment.r
```diff
@@ -44,3 +44,23 @@ init_shadowenv <- function() {
...
+            # attribute indicating the repos was set by us...
+            attr(r, 'irkernel') <- TRUE
+            options(repos = r)
```
flying-sheep: or in less lines:

```r
r[['CRAN']] = 'https://cran.r-project.org'
options(repos = structure(r, irkernel = TRUE))
```

R/environment.r
flying-sheep: where is this being called?

R/environment.r
```diff
@@ -44,3 +44,23 @@ init_shadowenv <- function() {
         stop("'edit()' not yet supported in the Jupyter R kernel")
     })
 }
+
+
+init_cran_repo <- function() {
+    local({
```
flying-sheep: this makes no sense, functions already are closures

flying-sheep:
OK, i better rewrite this completely instead of using 1000 comments:

```r
init_cran_repo <- function() {
    r <- getOption('repos')
    is_configured_by_user <- identical(r, c(CRAN = '@CRAN@'))
    if (!is_configured_by_user) {
        # the default repo according to https://cran.r-project.org/mirrors.html
        # uses geo-redirects
        r[['CRAN']] <- 'https://cran.r-project.org'
        # attribute indicating the repos was set by us...
        options(repos = structure(r, irkernel = TRUE))
    }
}
```

cran-comments.md
janschulz: done

R/environment.r
janschulz: done

R/environment.r
janschulz: nowhere... yet... done...

R/environment.r
```diff
@@ -44,3 +44,23 @@ init_shadowenv <- function() {
...
+        r <- getOption('repos');
+        is_configured_by_user <- !is.null(r) && identical(r, '@CRAN@')
```
flying-sheep: the null check is superfluous.

please just copy&paste [this version](https://github.com/IRkernel/IRkernel/pull/339#issuecomment-220296012)

R/execution.r
```diff
@@ -320,7 +320,11 @@ initialize = function(...) {
     # Create the shadow env here and detach it finalize
     # so it's available for the whole lifetime of the kernel.
     attach(NULL, name = 'jupyter:irkernel')
+
+    # Add stuff to the user environment and configure a few options
+    # in the currnet session
```
flying-sheep: “current”

R/environment.r
janschulz: an there I thought I managed to not remove all style problems ... done

R/environment.r
janschulz: I like it that I have a single line just for that so I can comment it... :-)

R/environment.r
janschulz: if you say so. :-) I thought so, too, but RStudio can't make so many things wrong... but this seems to be in since the beginning, so removing it... done

R/environment.r
```diff
@@ -44,3 +44,20 @@ init_shadowenv <- function() {
         stop("'edit()' not yet supported in the Jupyter R kernel")
     })
 }
+
+init_cran_repo <- function() {
+    r <- getOption('repos')
+    is_configured_by_user <- !identical(r, c(CRAN = '@CRAN@'))
+    if (!is_configured_by_user) {
```
flying-sheep: i think it’s a bit confusing if you invert twice.

how about

```r
is_default <- identical(r, c(CRAN = '@CRAN@'))
if (is_default) { ... }
```

or

```r
is_configured_by_user <- !identical(r, c(CRAN = '@CRAN@'))
if (is_configured_by_user) return()
...
```

R/environment.r
```diff
@@ -44,3 +44,20 @@ init_shadowenv <- function() {
...
+        # the default repo according to https://cran.r-project.org/mirrors.html
+        # uses geo-redirects
+        r['CRAN'] <- 'https://cran.r-project.org'
```
flying-sheep: [[]]

janschulz:
I addressed all comments, but did retain my way to do the test and set the repo options.

R/environment.r
janschulz: I don't agree: at least to me it's more meaningful to name it `is_configured_by_user` instead of  ìs_default. Maybe it could be "is_useable_mirror', but not `is_default` (which I would bet on is changed in one of the next versions to that mirror as well :-))

R/environment.r
janschulz: I tried to understand why that is different, but couldn't find anything. I tried it out and it didn't work:

```R
> r = getOption("repos")
> r
                        CRAN 
"https://cran.r-project.org" 
> r["CRAN"] <- "whatever"
> r
      CRAN 
"whatever" 
> r[["CRAN"] <- "whatever2"
Error: unexpected assignment in "r[["CRAN"] <-"
```

flying-sheep:
you missed the one about using `[[]]` instead of `[]`. `foo[1]` means the same as `foo[0:1]` means in python. in this case it makes no difference but it’s important to always do it right because it can cause subtle bugs.

i still think that two negations are confusing, but i can see that you think `is_configured_by_user` is clearer than `is_default`.

janschulz:
re `[[]]` vs `[]`:

```R
> r = getOption("repos")
> r
                        CRAN 
"https://cran.r-project.org" 
> r["CRAN"] <- "whatever"
> r
      CRAN 
"whatever" 
> r[["CRAN"] <- "whatever2"
Error: unexpected assignment in "r[["CRAN"] <-"
```

janschulz:
Ok, foudn my problem and changed to `[[]]`

janschulz:
removed the two negations by `is_unuseable_mirror`

janschulz:
Oh my, for this few new lines, I certainly produced the need for a lot of comments...

flying-sheep:
i guess that’s because it’s in parts copied from RStudio and we have a better code quality here :wink:

Issue #336, opened 2016-05-15
conversion failure


methodds:
When trying to display output in german language, I receive the following errors within the notebook:

````
plotQuote(thought$docs[[2]], width=120, text.cex=0.5, main="Topic 35")



Warning message:
In text.default(2.5, start + numlines[j - 1]/4 + numlines[j]/4, : conversion failure on 'von Moslems dominierten Stadtgebieten gewarnt. „Die Frage ist_ ob es tatsächlich sinnvoll ist_ sich in Problemvierteln_' in 'mbcsToSbcs': dot substituted for <e2>Warning message:
In text.default(2.5, start + numlines[j - 1]/4 + numlines[j]/4, : conversion failure on 'von Moslems dominierten Stadtgebieten gewarnt. „Die Frage ist_ ob es tatsächlich sinnvoll ist_ sich in Problemvierteln_' in 'mbcsToSbcs': dot substituted for <80>Warning message:
In text.default(2.5, start + numlines[j - 1]/4 + numlines[j]/4, : conversion failure on 'von Moslems dominierten Stadtgebieten gewarnt. „Die Frage ist_ ob es tatsächlich sinnvoll ist_ sich in Problemvierteln_' in 'mbcsToSbcs': dot substituted for <9e>Warning message:
In text.default(2.5, start + numlines[j - 1]/4 + numlines[j]/4, : conversion failure on 'in Vierteln mit einem hohen muslimischen Anteil_ als Jude durch das Tragen der zu erkennen gibt“_ sagte er dem RBB-' in 'mbcsToSbcs': dot substituted for <e2>Warning message:
In text.default(2.5, start + numlines[j - 1]/4 + numlines[j]/4, : conversion failure on 'in Vierteln mit einem hohen muslimischen Anteil_ als Jude durch das Tragen der zu erkennen gibt“_ sagte er dem RBB-' in 'mbcsToSbcs': dot substituted for <80>Warning message:
In text.default(2.5, start + numlines[j - 1]/4 + numlines[j]/4, : conversion failure on 'in Vierteln mit einem hohen muslimischen Anteil_ als Jude durch das Tragen der zu erkennen gibt“_ sagte er dem RBB-' in 'mbcsToSbcs': dot substituted for <9c>Warning message:
[...]

PR #340, opened 2016-05-19
added null device

flying-sheep:
circumvents #336 and friends

note that the device is still servicable using e.g. `dev.new('filename.png')`, as i used the null file as default argument.

### DIFF
--- a/R/environment.r
+++ b/R/environment.r
@@ -61,3 +61,13 @@ init_cran_repo <- function() {
 init_session <- function() {
     init_cran_repo()
 }
+
+
+init_null_device <- function() {
+    null_file <- switch(.Platform$OS.type, windows = 'NUL', unix = '/dev/null')
+    null_device <- function(filename = null_file, ...) png(filename, ...)
+    
+    if (identical(getOption('device'), pdf)) {
+        options(device = null_device)
+    }
+}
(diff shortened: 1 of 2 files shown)


flying-sheep:
@janschulz:

could you test if `plot(main = '\u2014')`

1. creates a warning without the PR
2. doesn’t create a warning with the PR

on windows?

flying-sheep:
maybe we can add a unit test for this. ideas?

janschulz:
If so, we nedd to do it in the python test code...

R/environment.r
```diff
@@ -61,3 +61,13 @@ init_cran_repo <- function() {
 init_session <- function() {
     init_cran_repo()
 }
+
+
+init_null_device <- function() {
+    null_file <- switch(.Platform$OS.type, windows = 'NUL', unix = '/dev/null')
```
janschulz: Mac OS X is included here?

janschulz:
```
init_null_device : null_device: no visible global function definition
  for ‘png’
init_null_device: no visible binding for global variable ‘pdf’
Undefined global functions or variables:
  pdf png
Consider adding
  importFrom("grDevices", "pdf", "png")
to your NAMESPACE file.
```

R/environment.r
flying-sheep:
```r
> help(.Platform)
OS.type: character string, giving the *O*perating *S*ystem (family) of
          the computer.  One of ‘"unix"’ or ‘"windows"’.
```

flying-sheep:
fixed, sorry

PR #345, opened 2016-05-24
Structured travis again

flying-sheep:
I think this concept works well.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+
+# NOTE:
+# This file uses some functions that have their bodies in subshells: (...)
+# This way they can use set -e, so the whole function returns a fail value once a command fails
+# Functions that don’t need this and/or export variables use the more common braces: {...}
+
+headline() {
+    echo -e "\033[33;1m$1\033[0m"
+}
+
+fold_wrap() {
+    local header=$1; shift
+    local cmd=$1; shift
+    
+    travis_fold start $cmd
+    headline "$header"
+    $cmd
+    local result=$?
+    travis_fold end $cmd
+    return $result
+}
+
+### BEFORE_SCRIPT ###
+
+install_r_packages() (
+    set -ex
+    
+    # Cairo was missing on OSX, no idea why this wasn't installed :-( Needed for repr
+    if [[ "$TRAVIS_OS_NAME" == osx ]]; then
+        Rscript -e "install.packages('Cairo')"
+    fi
+    
+    # Also test newer versions so that we catch errors earlier
+    if [[ "$DEPS" == github ]]; then
+        Rscript -e 'devtools::install_github(c("snoweye/pbdZMQ", "irkernel/repr", "irkernel/IRdisplay"))'
+    fi
+)
+
+create_conda_environment() (
+    set -ex
+    
+    # http://conda.pydata.org/docs/travis.html
+    if [[ "$TRAVIS_OS_NAME" == osx ]]; then os=MacOSX; else os=Linux; fi
+    wget "https://repo.continuum.io/miniconda/Miniconda3-latest-$os-x86_64.sh" -O miniconda.sh
+    bash miniconda.sh -b -p "$HOME/miniconda"
+    rm miniconda.sh # needed to prevent a error in the R CMD check part
+    
+    hash -r
+    conda config --set always_yes yes --set changeps1 no
+    conda update -q conda
+    # Useful for debugging any issues with conda
+    conda info -a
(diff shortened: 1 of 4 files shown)


.travis.sh
```diff
@@ -0,0 +1,88 @@
...
+    set -ex
+    
+    # Cairo was missing on OSX, no idea why this wasn't installed :-( Needed for repr
+    if [[ "$TRAVIS_OS_NAME" == osx ]]; then
+        Rscript -e "install.packages('Cairo')"
+    fi
+    
+    # Also test newer versions so that we catch errors earlier
+    if [[ "$DEPS" == github ]]; then
+        Rscript -e 'devtools::install_github("snoweye/pbdZMQ", "irkernel/repr", "irkernel/IRdisplay")'
```
janschulz: does that need a `c(...)` around the packages?

flying-sheep:
btw: we pulled @janschulz’ PR a bit too fast:

there’s a bug as the env doesn’t work anymore.

1. you wrote `envs`, which doesn’t exist
2. even with the correct `env`, i only see “no environment variables set” [e.g.](https://travis-ci.org/IRkernel/IRkernel/builds/132562591)

.travis.sh
```diff
@@ -0,0 +1,88 @@
...
+    bash miniconda.sh -b -p "$HOME/miniconda"
+    rm miniconda.sh # needed to prevent a error in the R CMD check part
+)
+
+configure_conda() (
+    set -ex
+    
+    hash -r
+    conda config --set always_yes yes --set changeps1 no
+    conda update -q conda
```
janschulz: IMO these two belong together in a "create_conda_environment". If there is a split then IMO between `conda info` and `conda create` (e.g. miniconda is not always useable without the `conda update conda` part)

.travis.sh
flying-sheep: good eye, yes!

but we didn’t see it since you broke DEPS anyway :laughing:

.travis.sh
flying-sheep: OK, why not!

flying-sheep:
huh, why does this fail for me and not you?

https://travis-ci.org/IRkernel/IRkernel/jobs/132565581#L852-L856

we always used “attach” here

janschulz:
```
grep -q 'WARNING' "$CHECK_LOG" ; [ $? -ne 0 ]
```

Means that the grep returns a non-0 exit code and the `set -e` errors out. Probably needs a `&& false || true`` to reverse it? (don't take my word for bash stuff...)

flying-sheep:
ah OK, makes sense. see, this works well!

flying-sheep:
Ah, nais, seems to work, and I even found a log checking bug!

flying-sheep:
hmm, doesn’t work on OSX. what did i change here?

https://travis-ci.org/IRkernel/IRkernel/jobs/132569514#L187

janschulz:
I think this is a problem with the order of the build which was masked due to the package caching: when I put "our" devtools installs in `before_install`, the old cache had already a devtools installed and therefore didn't complain. But now somehow the cache is gone and the real order is only installing the devtools in install: https://github.com/jimhester/travis-build/blob/8fcfe49c2c4102b67103c6f3f048ae7e7bd317fd/lib/travis/build/script/r.rb#L187

-> IMO the fix is changing `before_install` to `before_script`?

flying-sheep:
let’s try

.travis.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+# https://github.com/travis-ci/travis-rubies
```
janschulz: Please add a comment about `()` vs `{}` here :-)

```
Some functions are declared using `func_name() (...)` and some  `func_name() { }`. This is because we need ...
```

Otherwise I know I will change such stuff in a few month when I don't remember it...

flying-sheep: good point. i’ll also explain what set does and so on. i already documented the `{}` functions (they export stuff)

flying-sheep:
it worked! everything is documented and seems to be ready to go

flying-sheep:
all green. any objections to merge this?

takluyver:
Looks good to me. :-) @janschulz , want to push the button?

janschulz:
merge away :-)

@flying-sheep thanks for finding and fixing the envs/env error!

flying-sheep:
np. that’s why it’s so cool we’re three people. someone is always willing or able to do something :smile:

takluyver:
The three kernelteers! :-P

flying-sheep:
:laughing:

PR #352, opened 2016-06-10
Add basic support for store_history execution counts and request metadata

jrshust:
The jupyter message protocol states that you can avoid execution count incrementing even with silent false by setting store_history to false (https://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-counter-prompt-number).

Also, the IRkernel currently throws away top-level message metadata. I would like it to respect and return that.

Also made one small bug fix to the is_silent() call.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -160,15 +160,18 @@ handle_display_error = function(e) {
     send_error_msg(msg)
 },
 
-handle_value = function(obj) {
+handle_value = function(obj, visible) {
     log_debug('Value output...')
-    mimebundle <- prepare_mimebundle_kernel(obj, .self$handle_display_error)
-    if (length(intersect(class(obj), getOption('jupyter.pager_classes'))) > 0) {
-        log_debug('Showing pager: %s', paste(capture.output(str(mimebundle$data)), collapse = '\n'))
-        page(mimebundle)
-    } else {
-        log_debug('Sending display_data: %s', paste(capture.output(str(mimebundle$data)), collapse = '\n'))
-        send_response('display_data', current_request, 'iopub', mimebundle)
+    set_last_value(obj)
+    if (visible) {
+        mimebundle <- prepare_mimebundle_kernel(obj, .self$handle_display_error)
+        if (length(intersect(class(obj), getOption('jupyter.pager_classes'))) > 0) {
+            log_debug('Showing pager: %s', paste(capture.output(str(mimebundle$data)), collapse = '\n'))
+            page(mimebundle)
+        } else {
+            log_debug('Sending display_data: %s', paste(capture.output(str(mimebundle$data)), collapse = '\n'))
+            send_response('display_data', current_request, 'iopub', mimebundle)
+        }
     }
 },
 
--- a/.travis.sh
+++ b/.travis.sh
@@ -79,8 +79,10 @@ check_package() (
     # wc -l on OSX has some there...
     LINES=$(grep -v '* .*$' "$CHECK_LOG" | wc -l | sed -e 's/^[[:space:]]*//')
     echo "Lines: $LINES"
-    #  4 lines from the "attach" NOTE, 1 lines from the "Status" at the end
-    if [[ "_$LINES" != _5 ]]; then grep -v '* .*$' "$CHECK_LOG"; false; fi
+    #  4 lines from the "attach" NOTE,
+    #  4 lines from the ".Last.value" NOTE,
+    #  1 lines from the "Status" at the end
+    if [[ "_$LINES" != _9 ]]; then grep -v '* .*$' "$CHECK_LOG"; false; fi
 )
 
 test_kernel() (
--- a/R/utils.r
+++ b/R/utils.r
@@ -29,3 +29,9 @@ fromRawJSON <- function(r) {
     Encoding(s) <- 'UTF-8'
     fromJSON(s)
 }
+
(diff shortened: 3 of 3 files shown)


jrshust:
It seems all my tests fail because the published jupyter_kernel_test is quite ancient :-/. Will comment on that project to see what is up.

R/execution.r
```diff
@@ -78,11 +78,17 @@ Executor <- setRefClass(
...
 },
+is.store_history = function() {
+    current_request$content$store_history
+},
+request_metadata.get = function() {
+    current_request$metadata 
+},
 
 send_error_msg = function(msg) {
-    if (!is_silent()) return()
```
janschulz: Autsch

R/execution.r
```diff
@@ -78,11 +78,17 @@ Executor <- setRefClass(
 is.silent = function() {
     current_request$content$silent
 },
+is.store_history = function() {
+    current_request$content$store_history
+},
+request_metadata.get = function() {
```
janschulz: I see what you are trying to do here, but I really dislike the mixing of `_` and `.`.

`should.store.history` and `get.request.metadata` ?

R/kernel.r
```diff
@@ -104,10 +104,14 @@ new_reply = function(msg_type, parent_msg) {
         metadata      = namedlist())
 },
 
-send_response = function(msg_type, parent_msg, socket_name, content) {
+send_response = function(msg_type, parent_msg, socket_name, content, metadata = NULL) {
```
janschulz: When would you set metadata to something else than `request_metadata.get()`?

IMO there are tow "shorter" ways to structure this PR:

* define the function as `send_response = function(msg_type, parent_msg, socket_name, content, metadata = request_metadata.get())` and remove all the above `request_metadata.get()` calls
* If it is just a echoing back of the metadata don't redefine the function and just call `request_metadata.get()` in the `send_response` body.

janschulz:
I generally like this addition, and I understand the history stuff, but currently I don't understand the metadata stuff (why do we need to echo the old metadata?)

Therefore I would like to see the two main things separated into two commits with a proper commit message so that later one can see why these functions / functionality was added -> What's the underlying problem for each commit (Do we misinterprete the messaging spec? -> Link to the right section) and maybe if that's necessary: what's to be done about it and why is that the right remedy.

```
git reset HEAD~2
git add -p # only add the `is_silent -> is.silent hunk
git commit -m "Fix typo in is.silent function call"
# make the name changes in the code
# make more changes or discard stuff -> see questions above :-)
git add -p # -> Add only the hunks which belong to the "history" commit
git commit # add a proper commit message explaining the *why* of the commit
git add . # adds the rest
git commit 
```

Sorry to be pedantic here, but changes to the messaging spec implementation make me nervous :-)

man/Kernel-class.Rd
```diff
@@ -23,7 +23,7 @@ Has methods able to connect and talk to a Jupyter server.
 
 \item{\code{new_reply(msg_type, parent_msg)}}{Prepare a reply}
 
-\item{\code{send_response(msg_type, parent_msg, socket_name, content)}}{Send a response}
+\item{\code{send_response(msg_type, parent_msg, socket_name, content, metadata = NULL)}}{Send a response}
```
janschulz: This is wrongly in the "add a test case" commit :-/

takluyver:
I'm pretty sure that echoing all the metadata from a request in the reply is not the right thing to do. What was this trying to achieve?

R/execution.r
flying-sheep: maybe just underscores if we want to switch to them for everything.

PS: in the options, the mixing of underscores and dots exists, because we use `jupyter.` to mean a “namespace” for the options.

janschulz: I wouldn't mind :-) but `is.silent` should then also be changed... :-)

flying-sheep: Sure!

jrshust: Ok so it sounds like I should update to is_silent and should_store_history?

jrshust:
@takluyver I think you are correct on this one. My goal was to be able to tag a message with arbitrary metadata that would be propagated to all resulting messages (basically anything with the same parent_id). However, this does not seem to be part of the JMP contract so I can remove that one from the commit. 

Should I create a separate PR for the store_history addition then?

R/execution.r
jrshust: Yeah but it sounds like people might want to move to an underscore naming scheme anyway :D.

man/Kernel-class.Rd
jrshust: Hi @janschulz . I think I made a mistake with how I understood the metadata aspect within the JMP contract, so I will remove that and we can just proceed with the store_history commit.

R/execution.r
flying-sheep: yes, sorry for the silence!

man/Kernel-class.Rd
flying-sheep: that would be cool!

janschulz:
> Should I create a separate PR for the store_history addition then?

yes please :-)

flying-sheep:
hi @jrshust do you still plan to do the store_history thing? it would be cool to have it in!

jrshust:
Yes I will open a new PR

jrshust:
@flying-sheep See #407. 

Thanks!

PR #368, opened 2016-06-30
Use prepare_mimebundle from IRdisplay

flying-sheep:
i noticed that IRkernel doesn’t actually use IRdisplay’s machinery. i changed that, and now the tests pass.

depends on IRkernel/IRdisplay#33


flying-sheep:
OK, great. all but the weird mac fails are green, and this is only internal refactoring.

PR #370, opened 2016-06-30
Expose logging utilities to user env to enable upstream projects to use it

### LINKED ISSUE #369: Export logging utilities to enable usage in upstream projects
* We are working on improving logging/error message handling on the kernel side of declarative widgets (https://github.com/jupyter-incubator/declarativewidgets/pull/419).
* Instead of bringing in another logging package we would like to use what is already in IRkernel.  The [jupyter_options](https://github.com/IRkernel/IRkernel/blob/master/NAMESPACE#L5) for controlling the log level and log file are exported (so I can do `options(jupyter.log_level = 2L)`), but not the actual logging utilities [log_(debug|info|error)](https://github.com/IRkernel/IRkernel/blob/master/R/logging.r#L5).  Am I missing something or can I open a PR to export these?

poplav:
Fixes #369 which will be used in https://github.com/jupyter-incubator/declarativewidgets/issues/405
* Exposes the logging utility functions to the user env under the name `.irk_log_(debug|info|error)`


flying-sheep:
the failed check isn’t your fault but ours, sorry. we need to merge #368 to fix that

flying-sheep:
the `@include` is unnecessary.

it means that that file needs to be evaluated before this one can be evaluated, which isn’t true.

poplav:
@flying-sheep Updating the commit for the `@include` issue, thanks for clarifying.

flying-sheep:
OK, great!

Two more things:

1. The comment indicates that this is there to stay. As discussed a few minutes ago, please change it so it sounds less permanent
2. This is how the commit message looks like. When merging, I’ll shorten it and replace the :-code with its unicode equivalent `🔓`

	> `:unlock:, Expose logging utilities to user env to enable upstream pro…`

	Please use unicode in the future, as only Github interprets the codes

poplav:
@flying-sheep updated

R/environment.r
```diff
@@ -43,6 +43,12 @@ init_shadowenv <- function() {
     add_to_user_searchpath('edit', function(...) {
         stop("'edit()' not yet supported in the Jupyter R kernel")
     })
+
+    # Expose logging utilities for upstream projects
```
flying-sheep: this comment still suggests this is a reliable interface

poplav: @flying-sheep , hopefully the latest commit does it haha.  Thanks!

janschulz:
I don't really like to expose the main method together with the logging facilities so IMO this is not a temp workaround, but a rather permanent one. IMO the main method and the functions which should be available in the user session should be separated, so that a user doesn't have to import the package. IMO a package depending on jupyter should test if the kernel is running and only then use these function, not blindly use `IRkernel::something()`. 

We currently don't expose the package in the user session:

```R
env <- globalenv()
while (!is.null(env)) {
    cat(environmentName(env))
    cat("\n")
    env <- parent.env(env)
}
```
-> 
```
R_GlobalEnv
jupyter:irkernel
package:stats
package:graphics
package:grDevices
package:utils
package:datasets
package:methods
Autoloads
base
R_EmptyEnv
```

janschulz:
BTW: what is the difference between getting the `comm_manager` instance and this logging functions? The function to get the comm_manager is [currently exported](https://github.com/IRkernel/IRkernel/blob/master/R/kernel.r#L364-L369), so IMO if this is used in similar ways, then this should be handled similarly. 

Reading the documentation, you basically do
* start notebook
* run `library("declarativewidgets"); initWidgets()`
* load IRdisplay
* let the user use `display_html()` with the (html code) declaration of the widget

Is that right?

If so we should decide here what is the right way: getting the logging functions (and the comm_manager) from the user facing environment or by importing the IRkernel package.

Context seems to be https://github.com/poplav/declarativewidgets/blob/b355ac36753474a133a5d47e5d8e3891bea9d652/kernel-r/declarativewidgets/R/widget.r#L112-L117 which seems to be not using the exported method directly after importing IRkernel.

IMO opinion use the way via the user facing env:

```R
if (options("jupyter.is_running") {
   stop("Not running in a Jupyter R kernel")
cm <- .irk.get_comm_manager()
.irk.log_debug("...")
```

IMO its nice if the only thing exported from IRkernel is the main function and the kernelspec installer.  The log and the kernel manager rely on being inside a running R kernel, so just using the stuff from the IRkernel will lead to errors if users try that in a plain R session.

flying-sheep:
> BTW: what is the difference between getting the comm_manager instance and this logging functions? The function to get the comm_manager is currently exported, so IMO if this is used in similar ways, then this should be handled similarly. 

yeah, i think so too.

either a kernel is running, then you can use that function, or none is, then it returns NULL. there will never be multiple kernels available in R since user code always runs in `.GlobalEnv`

i think other functions could work the same.

flying-sheep:
yes. @takluyver @janschulz will we keep the names of the logging functions?

if so you can simply use `IRkernel:::log_debug` for now and drop the third colon once they’re exported

takluyver:
I imagine those names are OK, I haven't thought about it much.

Logging systems often have an 'info' level, which is higher than debug but lower than warning. We may add that at some point.

janschulz:
I like separating the two spaces: "user facing" (main and installer) and "package extension facing" (comm manager, logging), so I wouldn't mind having function names with dots for the latter.

On question:
what happens if we sometime in the future change the logging so that it uses the kernel itself instead of simply writing to stdout? In that case we would need to have access to the kernel object -> might that make a case for exporting the functions in the running kernel and not via package space? or will that simply do a `get_running_kernel().log_somthing()`?

flying-sheep:
> so I wouldn't mind having function names with dots for the latter.

yeah, if we don’t want to introduce yet another package to separate them :wink: 

> might that make a case for exporting the functions in the running kernel and not via package space?

i talked about this above: everything this package will ever do except `installspec` and `main` makes only sense with a running kernel.

and since user code will always rely on one singular `.GlobalEnv` (per R session), there’s no way this package will some day change to allow multiple kernels running in parallel.

janschulz:
> and since user code will always rely on one singular .GlobalEnv (per R session), there’s no way this package will some day change to allow multiple kernels running in parallel (in one R session).

I'm more concerned about *no* kernel running but the user doing a `library(IRkernel); log_debug()`. Currently all such function "work" (getting the comm is NULL and logs write to stdout/err) but what happens if we gain a functions which depends on a running kernel (-> is `get_running_kernel().log_somthing()` enough)?

flying-sheep:
ah, gotcha. hmm.

we could either do a runtime check for the “in_kernel” option or store the `.stderror` var in some env and only set it to a valid value once the kernel runs.

poplav:
The commit has been reverted to exporting the logging utilities instead of adding them to the user search path.

flying-sheep:
cool! as the test results indicate, you forgot to create the .Rd files.

1. use `#' @name log_*` so that the functions end up on the same help page (you only need to write description text for the first one, and need only `@name` and `@export` on the other functions)
2. in the description text please mention configuring log level and ~file via [option](https://github.com/IRkernel/IRkernel/blob/8cdfe5e001e6ef746babcdfea5a6bf1692df9678/R/options.r#L8-L9) and [env variable](https://github.com/IRkernel/IRkernel/blob/8cdfe5e001e6ef746babcdfea5a6bf1692df9678/R/options.r#L19-L20)
3. use roxygen2 to create the .Rd files?

poplav:
@flying-sheep documentation was updated

R/logging.r
```diff
@@ -2,18 +2,29 @@
 # Per default, only error messages are shown
 # levels: 3 = DEBUG, 2 = INFO/MSG, 1 = ERROR
 
+#' A set of exported logging utilities that have the capability to be used in upstream projects
+#' Log level and log file can be set via R package options e.x. options(jupyter.log_level = 2L))
```
flying-sheep: > …projects

“.” missing

> e.x.

e.g.

> `options(jupyter.log_level = 2L))`

`\code{options(jupyter.log_level = 2L))}`

R/logging.r
```diff
@@ -2,18 +2,29 @@
...
+#' or from the environment variables JUPYTER_LOG_LEVEL and JUPYTER_LOGFILE
+#'
+#' @name log_*
+#' @param \dots message to log
```
flying-sheep: please do `...`, i think `R cmd check` can’t interpret this.

also there need to be two spaces after the parameter name, i.e. `...  message to log`

flying-sheep:
that’s almost perfect!

once you address the line comments add a title like this, i’ll merge it:

```r
#' Kernel logging functions
#' 
#' A set of exported logging…
```

poplav:
@flying-sheep, let me know if it's perfect now.

flying-sheep:
yes, pretty much :smile:

PR #371, opened 2016-07-01
integrate kernel tests in testthat

flying-sheep:
and updated travis with it

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -70,8 +70,6 @@ build_package() {
 check_package() (
     set -ex
     
-    # we don't have test_that tests yet.
-    Rscript -e 'devtools::test(reporter="check")'
     R CMD check "$PKG_TARBALL" --as-cran
     ! grep -q 'WARNING' "$CHECK_LOG"
     # .. because ': ' was resulting in an replacement by travis and an error
@@ -91,5 +89,5 @@ test_kernel() (
     R CMD INSTALL "$PKG_TARBALL"
     Rscript -e 'IRkernel::installspec()'
     Rscript -e 'IRkernel::installspec(name = "testir", displayname = "testir")'
-    python -W ignore::DeprecationWarning test_ir.py
+    Rscript -e 'devtools::test()'
 )
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,8 +25,7 @@ Imports:
     jsonlite (>= 0.9.6),
     uuid,
     digest
-Remotes:
-  IRKernel/IRdisplay,
+Remotes: IRKernel/IRdisplay,
 Collate:
     'logging.r'
     'comm_manager.r'
--- a/IRkernel.Rproj
+++ b/IRkernel.Rproj
@@ -12,6 +12,8 @@ Encoding: UTF-8
 RnwWeave: Sweave
 LaTeX: pdfLaTeX
 
+AutoAppendNewline: Yes
+
 BuildType: Package
 PackageUseDevtools: Yes
 PackageInstallArgs: --no-multiarch --with-keep.source
--- a/tests/testthat/test_ir.py
+++ b/tests/testthat/test_ir.py
@@ -9,6 +9,7 @@ options(jupyter.rich_display = FALSE)
 {}
 options(jupyter.rich_display = TRUE)
 '''
+#this will not work!
 #withr::with_options(list(jupyter.rich_display = FALSE), {})
 
 
@@ -63,6 +64,7 @@ class IRkernelTests(jkt.KernelTests):
     ]
 
     def test_display_vector(self):
+        """display of vectors"""
         code = '1:3'
         reply, output_msgs = self._execute_code(code)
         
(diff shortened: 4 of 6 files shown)


takluyver:
My take is that this is more complexity than it's worth - we're can already run the two sets of tests on Travis without integrating them, and for manual testing I'd rather run two separate straightforward test commands than debug odd cross-language issues. If you want to maintain it, however, that's your lookout. ;-)

flying-sheep:
what cross-language issues do you mean?

i only communicate using strings and dicts of string, which rPython can handle well.

janschulz:
I think a hint is the commit text: "Work around rPython’s insanity" :-)

flying-sheep:
the insanity being that it’s compiled against a single python version, and there isn’t a rPython3 or so. it’s really stupid to have to globally decide which python version you want to use.

also i’m currently trying to keep the ancient broken readline library from interfering, which isn’t rPython’s fault, but conda’s :sob:

takluyver:
Yeah, that's the sort of cross language issues I mean. I'd rather just run the Python tests directly than deal with rPython.

If it's really necessary to implement them, why not just use R's subprocess machinery to run python? That seems less faff than doing it in process.

takluyver:
(I'm not saying it's rPython's 'fault' that this is tricky. I'm not interested in assigning blame. It just evidently is tricky, and I don't see any real advantage in dealing with it)

flying-sheep:
yeah, that would also work, but it would have to start up the python interpreter for each test…

and you can blame rPython. it’s obviously a quick hack that was never extended to be more. https://github.com/asieira/SnakeCharmR could be better but it’s not on CRAN

takluyver:
> yeah, that would also work, but it would have to start up the python interpreter for each test…

We can definitely run the whole Python test suite in one interpreter. If this is something to do with the number of tests it shows it's run... who cares? ;-)

flying-sheep:
what it currently does is converting the python unittest suite into a testthat suite, completely integrated. of course it’s overkill, but it’s also convenient.

takluyver:
I still think complete integration is not worth the complexity. Remember the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle)!

flying-sheep:
what kind of integration do you want to see?

i definitely want the kernel tests to be run via `devtools::test()`

takluyver:
If the priority is that the R tests run the Python tests, just run the
whole python test suite in a subprocess and check that it exits with 0.

tests/testthat/get-tests.py
```diff
@@ -0,0 +1,33 @@
...
+    for t_or_s in suite:
+        if isinstance(t_or_s, unittest.TestSuite):
+            yield from get_test_ids(t_or_s)
+        else:
+            yield t_or_s.id()
+
+module_suite = unittest.defaultTestLoader.loadTestsFromModule(test_ir)
+test_ids = list(get_test_ids(module_suite))
+
+# this has to be this compex since an individual TestCase does not call setUpClass
```
janschulz: s/compex/complex/

tests/testthat/test-kernel.r
```diff
@@ -0,0 +1,29 @@
+context('kernel')
+
+Sys.setenv(
+    PYTHONWARNINGS = 'ignore::DeprecationWarning',
+    PYTHONPATH = paste(getwd(), Sys.getenv('PYTHONPATH'), sep = ':'))
```
janschulz: This is wrong on all my R environments :-(

I currently code in RStudio and when starting RStudio, no PYTHONPATH is available. The default environment, which is in PATH also doesn't have the kernel testing tool installed, that is in an conda env which I have to activate first.

janschulz:
Another way to do the tests in R would be to create a kernel object and afterwards set the base display functions to something we control (like in IRdisplay) and then implement the (or some) tests in R.

flying-sheep:
very good idea!

flying-sheep:
OK, i switched to a simpler approach: make unittest output text and parse it from R

takluyver:
> make unittest output text and parse it from R

Why? What's wrong with just running it as a subprocess using [system2](https://stat.ethz.ch/R-manual/R-devel/library/base/html/system2.html) and asserting that the exit code is 0?

flying-sheep:
because i wanted it to be neat.

and since it’ll still fail the test if something happens that i didn’t think about, it’s also not a risk

takluyver:
But why add the extra complexity to write the test results in a different format and then parse them out again? All we need to do is get the output to the console, and check whether the test suite as a whole passed or failed. This should be simple, there's no need to come up with Heath Robinson solutions.

flying-sheep:
Integration. E.g. you can use runthat or other test runners.

takluyver:
But if the test is just 'run the Python tests, check that they passed', that should work under any test runner, right? It will only show up as one test rather than several, but that doesn't really matter.

> there’s no downsides.

Extra complexity is a downside by itself!

flying-sheep:
then sending everything but basic ZMQ functionality tests through the jupyter kernel tests interface is overkill as well.

takluyver:
There is a need for some integration tests that start a kernel and communicate with it, because that's the contract the kernel has with the Jupyter system as a whole. Once the machinery to do that is in place, there's little extra complexity involved in writing more tests using it. But sure, I'd be happy to have the bulk of the tests in R and relatively few in Python, so long as we're exercising the ZMQ interface properly.

If there's a good reason for adding complexity, we can. But we're not building random extra complexity on a whim.

This is getting rancorous, and it's dragged on long enough. I'm going to push to this branch to do what I mean.

takluyver:
Merged in #375

Issue #349, opened 2016-06-05
Investigate using svglite


flying-sheep:
[svglite](https://github.com/hadley/svglite) might be a device that

1. can handle `/dev/null` even on OSX (i think it uses C++ stdlib stuff, not Apple APIs, so that’s a valid hope) and therefore serves as dummy device
2. can be used in repr for SVG, invalidating the need to use the isolate flag in IRkernel (since svglite it’s built to produce embeddable SVGs)

if it can do both those things, we’ll simply make IRkernel hard depend on it, and repr prefer it to all other possible SVG devices (not a compromise/hack, because it *does* produce the best SVG code). then it will be always used by IRkernel, and we can cease using the isolate flag.

PR #372, opened 2016-07-01
Remove svg from the default sent image types

janschulz:
SVG currently has a few problems:

* complicated plots can lead to very big streams sent which takes ages to show up in the browser and lead to bad scrolling behaviour
* svg output in the browser is not saveable in chrome (as it is isolated into an iframe)
* using two plot formats also means that two plots have to be drawn which takes twice as long.

Closes: https://github.com/IRkernel/IRdisplay/issues/13

### DIFF
--- a/test_ir.py
+++ b/test_ir.py
@@ -88,18 +88,15 @@ class IRkernelTests(jkt.KernelTests):
         code = 'plot(1:3)'
         reply, output_msgs = self._execute_code(code)
         
-        # we currently send three formats: png, svg and text/plain
+        # we currently send two formats: png, and text/plain
         data = output_msgs[0]['content']['data']
-        self.assertEqual(len(data), 3, data.keys())
+        self.assertEqual(len(data), 2, data.keys())
         self.assertEqual(data['text/plain'], 'plot without title')
-        self.assertIn('image/svg+xml', data)
         self.assertIn('image/png', data)
         
-        # we isolate only svg plots
+        # we isolate only svg plots, which are not included in the default types
         metadata = output_msgs[0]['content']['metadata']
-        self.assertEqual(len(metadata), 1, metadata.keys())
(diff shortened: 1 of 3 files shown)


flying-sheep:
i’m OK with merging that, but please stay factual:

* not complicated, but *broken* (e.g. that huge ggplot object) or *useless* plots do. useless plots have e.g. huge amounts of points that are on top of each other and therefore and give no idea of density
* the fact that they’re not savable is 1. preventable (#349) and 2. not limited to SVG
* the time it takes is OK

the real problem is that other tech have too many bugs regarding SVG so it’s pragmatic to remove it.

flying-sheep:
jup, you need to adapt the tests.

janschulz:
Sorry for the missed (failing) tests, fixed now, hopefully.

I don't agree on point 1 and 3 but didn't want to start arguing again with you, so I just removed the commit message and pointed them here and the original issue.

test_ir.py
```diff
@@ -88,19 +88,16 @@ def test_irkernel_plots(self):
...
+        # we currently send two formats: png, and text/plain
         data = output_msgs[0]['content']['data']
-        self.assertEqual(len(data), 3, data.keys())
+        self.assertEqual(len(data), 2, data.keys())
         self.assertEqual(data['text/plain'], 'plot without title')
-        self.assertIn('image/svg+xml', data)
         self.assertIn('image/png', data)
         
-        # we isolate only svg plots
+        # we isolate only svg plots, which are not included anymore
```
flying-sheep: please add a test checking if they’re isolated when configured to be there

janschulz: done, sorry for missing that. :-/

janschulz:
so, finally got the errors out... Apart from the osx build, which fails in installing latex?

flying-sheep:
merged: 8cdfe5e001e6ef746babcdfea5a6bf1692df9678

PR #375, opened 2016-07-05
Simpler integration of kernel tests into R test framework

takluyver:
See discussion on #371


flying-sheep:
looks good to me!

PS: i’ll probably do another PR soon that also integrates the testing systems, but does so in a smarter way. i’ve crated a JSON TestRunner, and we can consume its output using like 5 lines via a loop and [`jsonlite::stream_in`](http://rpackages.ianhowson.com/cran/jsonlite/man/stream_in.html)

flying-sheep:
maybe you could remove the `DeprecationWarning`s in another commit. you can use `-W ignore::DeprecationWarning`

takluyver:
> i’ve crated a JSON TestRunner...

I still don't see any advantage to this extra complexity. Complexity in a package you've created is still complexity you have to maintain. Let's keep it simple, at least until we identify a clear need for something more.

> maybe you could remove the DeprecationWarnings in another commit...

Done. I'm not sure if it will work or not, as DeprecationWarnings are ignored by default, so something in the test machinery must be turning them on.

tests/testthat/test_kernel.r
```diff
@@ -0,0 +1,6 @@
+context('kernel')
+
+status <- system2('python3', 'test_ir.py', wait=TRUE)
```
janschulz: It will be interesting what will happen if we submit that to CRAN: on my system, there is no python3 and in RStudio (or my r package install script), I don't have the kernel testing tool installed, so this tests will basically always fail.

takluyver:
Suppressing the DeprecationWarnings does appear to have worked :-)

tests/testthat/test_kernel.r
takluyver: I guess we'll just remove this file from the tarball we submit to CRAN, in that case :-(

takluyver: ...or use a series of system commands to install Anaconda ;-). But it may well block network access.

flying-sheep:
> as DeprecationWarnings are ignored by default

they are in python, but they aren’t in tests. (i.e. python per default installs a warning filter, which is removed during unit tests)

tests/testthat/test_kernel.r
flying-sheep: we add `python3` and `jupyter_kernel_test` to `SystemRequirements` with an explanation that they’re needed for tests.

Issue #381, opened 2016-07-15
Replace dockerfile


poplav:
When trying to run the kernel in a docker container following the instructions on the readme I ran into https://github.com/jupyter/notebook/issues/1539.  
https://github.com/jupyter/notebook/issues/1514 was opened to remove and no longer support the jupyter notebook dockerfile, which is what the irkernel dockerfile is inheriting from.  They are instead opting to default to [jupyter/docker-stacks](https://github.com/jupyter/docker-stacks).

One proposal is:  Remove the Dockerfile and update the readme to simply point to the supported [jupyter/r-notebook](https://github.com/jupyter/docker-stacks/tree/master/r-notebook).  In that process I think we can also expand on the readme to include irkernel build instructions when in a dockerized environment: i.e. volume mount source, go into container, `R CMD INSTALL -l /opt/conda/lib/R/library .`  Or add a make target to run this container and reinstall the volume mounted source.  Thoughts?

cc'ing @benjaminabel as it looks like you authored the original dockerfile

PR #382, opened 2016-07-18
Make dockerfile use jupyter docker stacks

poplav:
Fixes https://github.com/IRkernel/IRkernel/issues/381:
* Adds `dev_image` and `dev` targets to the makefile
  * `dev_image` - builds/installs necessary IRkernel dependencies from the [jupyter docker-stacks/r-noteook](https://github.com/jupyter/docker-stacks/tree/master/r-notebook) base image
  * `dev` - mounts IRkernel source into the `dev_image` container, installs the the IRkernel package from the source, and starts the Jupyter notebook
* Removes old Dockerfile as it has been replaced.
* Makes the distinction between running a stable release version of the irkernel in docker vs running in dev mode in the readme


flying-sheep:
i have no idea how docker works, so someone else has to review this

takluyver:
> i have no idea how docker works

Magic! :sparkles: 

I'll take a look

Makefile
```diff
@@ -9,3 +13,19 @@ check:
...
 test: 
 	Rscript  -e "library(testthat); test()"
+
+dev_image:
+	@docker run -it --name irkernel-build \
+	$(DEV_IMAGE) bash -c 'Rscript -e "install.packages(c('\''devtools'\''), repos = c('\''http://irkernel.github.io/'\'', '\''http://cran.rstudio.com'\''))" && \
+	                        Rscript -e "library(devtools)" \
+	                        -e "install_github('\''IRkernel/repr'\'')" \
+	                        -e "install_github('\''IRkernel/IRdisplay'\'')"'
+	@docker commit irkernel-build $(DEV_IMAGE)
```
takluyver: Shouldn't this be committing with a different name, not overwriting `jupyter/r-notebook`? The user should still be able to run the stable version after they've run this, without the two interfering.

Makefile
```diff
@@ -9,3 +13,19 @@ check:
...
+	@docker rm -f irkernel-build
+
+dev:
```
takluyver: And shouldn't `dev` depend on `dev-image`?

poplav:
@takluyver, added a distinction between base/dev image and added the dependency.

takluyver:
Thanks, that looks better. I assume you've tested doing `make dev`?

I also think we should rename the Make targets so they're more clearly associated with docker, e.g. `docker_dev` and `docker_dev_image`.

Makefile
```diff
@@ -9,3 +14,19 @@ check:
...
 	Rscript  -e "library(testthat); test()"
+
+docker_dev_image:
+	@docker run -it --name irkernel-build \
+	$(BASE_IMAGE) bash -c 'Rscript -e "install.packages(c('\''devtools'\''), repos = c('\''http://irkernel.github.io/'\'', '\''http://cran.rstudio.com'\''))" && \
+	                        Rscript -e "library(devtools)" \
+	                        -e "install_github('\''IRkernel/repr'\'')" \
+	                        -e "install_github('\''IRkernel/IRdisplay'\'')"'
+	@docker commit irkernel-build $(DEV_IMAGE)
+	@docker rm -f irkernel-build
```
parente: With modern docker (1.9+?), it's better to contain these steps in a Dockerfile.dev and simply `docker build -f Dockerfile.dev -t $(DEV_IMAGE) .`  to get the added value of caching.

parente: https://github.com/jupyter-incubator/dashboards/blob/master/Makefile#L21

and

https://github.com/jupyter-incubator/dashboards/blob/master/Dockerfile.dev

for reference.

poplav:
@parente, updated and I did notice the caching.

takluyver:
Thanks!

PR #390, opened 2016-08-07
Don’t export the kernel

flying-sheep:
There can only be one instance that has to be running in a `while (TRUE)` loop, and there’s neither need nor an official way to access the instance variable from a running session.

This is just confusing IMHO


takluyver:
Seems fine to me.

flying-sheep:
OK, just wanted confirmation

PR #391, opened 2016-08-10
Added colorful logging

flying-sheep:
This will add the common dependency crayon.

shouldn’t be a problem, but maybe we should wait with new dependencies until having it on CRAN makes this trivial (i.e. we won’t have to mention dependencies in readmes anymore)


R/logging.r
```diff
@@ -10,26 +10,29 @@
 #'
 #' @name log_*
 #' @param ...  message to log
+#' @importFrom crayon green
 #' @export
 log_debug <- function(...) {
     if (isTRUE(getOption('jupyter.log_level') >= 3L)) {
-        log_msg('DEBUG', sprintf(...))
+        log_msg(green('DEBUG'), sprintf(...))
```
janschulz: IMO the coloring should be done in log_msg_stderror as as far as I can set will now output escape chars into the text based logfile.

janschulz:
I'm not sure this is helpfull: the R kernel log is currently mixed together  with the normal jupyter notebook log (at least on my windows cmd window which opens when the notebook is opened). But that log isn't colored so it becomes a mix of colored and uncolored log entries.

R/logging.r
flying-sheep: makes sense. i’ll simply add a parameter to that function

flying-sheep:
still helpful IMHO: better help visually distinguish a part of the logspam than nothing.

R/logging.r
```diff
@@ -59,14 +62,15 @@ log_msg_stderror <- function(lvl, msg){
...
 log_msg_logfile <- function(lvl, msg) {
     cur_logfile <- getOption('jupyter.logfile')
     if (!is.na(cur_logfile)) {
         if (.is_changed_logfile(cur_logfile)) {
-            log_msg_stderror('INFO', sprintf('Logging to %s', cur_logfile))
+            log_msg_stderror(blue('INFO'), sprintf('Logging to %s', cur_logfile))
         }
         log_con <- file(cur_logfile, open = 'ab')
-        writeBin(charToRaw(sprintf('%s %s: %s\n', format(Sys.time()), lvl, msg)), log_con, endian = 'little')
+        writeBin(charToRaw(sprintf('%s %s: %s\n', format(Sys.time()), strip_style(lvl), msg)), log_con, endian = 'little')
```
flying-sheep: i remove the styling now here

Issue #399, opened 2016-09-02
Dev Image broke due to missing crayon dependency not being installed


poplav:
There is a regression with the dockerized development environment following https://github.com/IRkernel/IRkernel/pull/391.  
Crayon a required dependency was not added to the Dockerfile as a dependency to install.  Thus the command `make docker_dev` fails with `ERROR: dependency ‘crayon’ is not available for package ‘IRkernel’`

I'll open a quick PR to fix this.

poplav:
Fixed by #400

poplav:
Fixed in docker stacks...closing.

flying-sheep:
thank you!

PR #400, opened 2016-09-02
Adds crayon as a dependency to be installed in the docker dev image

poplav:
Fixes https://github.com/IRkernel/IRkernel/issues/399

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,6 +2,6 @@
 
 FROM jupyter/r-notebook
 
-RUN Rscript -e "install.packages(c(\"devtools\"), repos = c(\"http://irkernel.github.io/\", \"http://cran.rstudio.com\"))"
+RUN Rscript -e "install.packages(c(\"devtools\", \"crayon\"), repos = c(\"http://irkernel.github.io/\", \"http://cran.rstudio.com\"))"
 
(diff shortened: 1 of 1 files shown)


flying-sheep:
thanks, but please fix it [here](https://github.com/jupyter/docker-stacks/blob/master/r-notebook/Dockerfile) for everyone!

poplav:
@flying-sheep I feel like it should only be fixed in the docker stacks when the irkernel version is updated.  At the current time the docker stacks serves its purpose as a stable environment.  The docker image here serves its purpose as a dockerized dev environment that should get the bleeding edge updates.  Once the base docker image is updated on the cycle of a new irkernel release I think the fixes here can be removed?  Thoughts?  In essence I am trying to say there is a difference between the stable env in docker stacks and the dev version here.

flying-sheep:
ah, but there isn’t a stable version of IRkernel out yet…

we’re still 0.x, and don’t even have it up on CRAN. but i’ll reopen

takluyver:
Calling projects 0.x doesn't really signify much nowadays, I think. IPython was in '0.x' for over ten years. It doesn't stop people treating the 0.x releases as releases :-)

flying-sheep:
would releasing 0.7 ASAP help :wink:

poplav:
I'll open a PR on docker stacks and close this...this shouldn't be an issue.

flying-sheep:
thanks! i just thought that spreading our dependencies and instructions over too many places, it gets unmaintainable

PR #401, opened 2016-09-02
Fix discovering/running of unit tests due to file extension issue

poplav:
Fixes #402

Issue #402, opened 2016-09-02
unit tests are not run from R CMD CHECK


poplav:
From the `testthat` docs under the section [integration with r cmd check](https://github.com/hadley/testthat#integration-with-r-cmd-check) it says the test files can be .r or .R, but the main `tests/testthat.R` must be with a capital .R not .r.  
Currently we have testthat.r and it is not running on the build in the [check package travis step](https://github.com/IRkernel/IRkernel/blob/master/.travis.sh#L70).

poplav:
Fixed by https://github.com/IRkernel/IRkernel/pull/401

flying-sheep:
`devtools::test()` should just run all .r/.R files in `/tests`… why doesn’t it?

poplav:
Yes, `devtools::test()` does run all the tests in [test_kernel](https://github.com/IRkernel/IRkernel/blob/master/.travis.sh#L86).  
Although, the unit level tests are not being run in [check_package](https://github.com/IRkernel/IRkernel/blob/master/.travis.sh#L70) when you do a `R CMD check`, confirmed via the logs and see [integration with r cmd check](https://github.com/hadley/testthat#integration-with-r-cmd-check).

I think there should be a clearer divide between unit and system tests as well and this PR addresses that.  The unit tests would now be run in `check_package` via `R CMD check` and the system tests would be run in `test_kernel` via directly running the python test file.

flying-sheep:
that’s just confusing. tests are tests.

flying-sheep:
@poplav what do you mean by “system tests”? we only have unittests.

do you mean that the kernels have to be installed for the `test_ir` script to run?

i think we should fix *that*, then. @takluyver is there a way to use `jupyter_kernel_test` with a kernel installed in a custom location (i.e. without running `installspec`)

flying-sheep:
what i want:  you should be able to run `R CMD CHECK IRkernel-0.7.tar.gz` with only the dependencies installed but no kernelspecs and have all tests run.

poplav:
I was making the distinction that `jupyter_kernel_test` is a system test as it requires external dependencies/is more integration.  The tests in testthat which get run in `R CMD check` should be treated as unit tests with no dependencies.  #401 seemed to address that distinction and pass.  I'll let you take over it in #404 thanks!

flying-sheep:
the unit tests are really basic though, so they aren’t really helpful.

[...]

PR #404, opened 2016-09-07
install, use, and remove testir kernel from tests

flying-sheep:
i’m reusing parts from #402

fixes #401


flying-sheep:
hey @takluyver, do you allow me to integrate the more fancy test integration now :wink:?

https://travis-ci.org/IRkernel/IRkernel/jobs/158155545#L866-L883

flying-sheep:
so i’m going to merge the more complex solution as it has tangible benefits:

1. a `R CMD check` or `devtools::check()` or `devtools::test()` will run *all* tests.

	a new contributor (and a lazy long time contributor) might only know/want to hit the RStudio shortcut for testing/checking once instead of learning/remembering that there are multiple test suites to run

2. `R CMD check` displays only the last 13 lines of testing output. testthat accounts for this and displays a nice summary in those last 13 lines.

	if we only check if the python test suite exited successfully, we lose testthat’s summary and won’t know which tests failed.

	this means that an alternative solution would 1. skip all python tests on travis 2. manually run the python test suite on travis. i want to avoid behavioral differences locally and on travis.

PR #406, opened 2016-09-10
bind last output to .Last.value

### LINKED ISSUE #405: .Last.value does not work in IRkernel
I guess the reason `.Last.value` doesn't work in IRkernel is exactly the same reason why it doesn't work in `knitr`: http://stackoverflow.com/questions/31475226/using-last-value-in-rmarkdown-knitr

With the understanding that it doesn't work, I do find that it is quite handy to have the last output stored so that it could be used later, it is particularly helpful when using the autograde feature of [nbgrader](https://github.com/jupyter/nbgrader).

Does it make sense and is it difficult to enable `.Last.value`?

randy3k:
close #405 

Thanks @flying-sheep for pointing out the relevant code. Now `.Last.value` works as expected (even within a cell).

screen shot 2016-09-11 at 12 27 43 am

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -79,8 +79,10 @@ check_package() (
     # wc -l on OSX has some there...
     LINES=$(grep -v '* .*$' "$CHECK_LOG" | wc -l | sed -e 's/^[[:space:]]*//')
     echo "Lines: $LINES"
-    #  4 lines from the "attach" NOTE, 1 lines from the "Status" at the end
-    if [[ "_$LINES" != _5 ]]; then grep -v '* .*$' "$CHECK_LOG"; false; fi
+    #  4 lines from the "attach" NOTE,
+    #  4 lines from the ".Last.value" NOTE,
+    #  1 lines from the "Status" at the end
+    if [[ "_$LINES" != _9 ]]; then grep -v '* .*$' "$CHECK_LOG"; false; fi
 )
 
 test_kernel() (
--- a/R/execution.r
+++ b/R/execution.r
@@ -160,15 +160,18 @@ handle_display_error = function(e) {
     send_error_msg(msg)
 },
 
-handle_value = function(obj) {
+handle_value = function(obj, visible) {
     log_debug('Value output...')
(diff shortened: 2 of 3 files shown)


randy3k:
I think the test failure is not relevant, but there is a note

```
Found the following possibly unsafe calls:
File ‘IRkernel/R/utils.r’:
  unlockBinding(".Last.value", .BaseNamespaceEnv)
```

Basically, it says unlocking `.Last.value` is not safe, but I don't know there is a way to do it safely. Perhaps assigning the last value to sth like `.jupyter.last_value`?

randy3k:
I just note that the behaviour is different from R console.

```
> x <- 1
> x
[1] 1
> .Last.value
[1] 1
> x <- 2
> .Last.value
[1] 2
```

screen shot 2016-09-10 at 4 31 46 pm

It is because `x <- 2` has no output...

flying-sheep:
well, [the docs say that it lives in the base namespace](http://www.rdocumentation.org/packages/base/versions/3.3.1/topics/Last.value), so you’re doing the right thing.

we’re basically replacing the frontend of the whole R runtime, so it’s expected that we do a few things that are considered “unsafe”.

and [of course your changes cause the test failure](https://travis-ci.org/IRkernel/IRkernel/jobs/159067956#L904): we just test for the number of lines the “NOTE” has… we could do this better, but for now we have to update and account for every single line. as you can see it went from 4→9. what lines gets added?

randy3k:
I have updated the tests, 6 out of 8 of the jobs pass except the devel versions on linux. I have no clues how to fix them..

flying-sheep:
it’s not “8 lines from the ‘attach’ NOTE”, though. it’s still 4 for that one and 4 for your new one.

randy3k:
I fixed the comments in `.travis.sh`

flying-sheep:
great. i’ll wait for the test results and merge

flying-sheep:
oh, and of course: thank you :smile:

randy3k:
😀
You are welcome

PR #407, opened 2016-09-13
Add store_history flag to determine when to increment execution count

jrshust:
Don't increment execution count if store history is set to false:
https://jupyter-client.readthedocs.io/en/latest/messaging.html#execution-counter-prompt-number

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -76,6 +76,10 @@ is_silent = function() {
     current_request$content$silent
 },
 
+should_store_history = function() {
+    current_request$content$store_history
+},
+
 send_error_msg = function(msg) {
     if (!is_silent()) return()
     send_response('stream', current_request, 'iopub',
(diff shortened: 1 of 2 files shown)


jrshust:
> The kernel should have a single, monotonically increasing counter of all execution requests that are made with store_history=True.

I have a project that does execute_requests in the background and the user ends up seeing execution counts going up by 2 instead of 1 since it's not currently respecting store_history: false (I can't set silent to true because I do stuff with the output).

flying-sheep:
uh, wait… you don’t actually do anything with the flags, so how do the tests pass?

jrshust:
I updated my PR to use the flags and made sure I ran the test correctly locally. It definitely failed w/o the flags and succeeded with them.

It is not clear to me how travis is running the tests, but I don't have much experience with travis. I don't see any output during the build to suggest the tests are run. Do we just need to add that to the travis yml?

flying-sheep:
thank you! :smiley:

PR #408, opened 2016-09-15
Dockerize running all the tests in the makefile when developing

poplav:
Off the heals of #402 being fixed.
Here is the dockerized version of running the tests, which should be more convenient.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -2,6 +2,10 @@
 
 FROM jupyter/r-notebook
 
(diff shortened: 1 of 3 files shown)


poplav:
Anybody,
alt text

flying-sheep:
i have no docker exp and a busted elbow, so @takluyver?

takluyver:
Looks reasonable, thanks for the reminder @poplav .

Issue #299, opened 2016-04-17
Add completion for function arguments


balta2ar:
In Python notebooks I use completion for function arguments all the time. Could you please add the same here?

flying-sheep:
what keyboard shortcut is that?

and if you know: what [message type](http://jupyter-client.readthedocs.org/en/latest/messaging.html) is it?

balta2ar:
The shortcut is `Shift-Tab`. Sorry, I don't know its message type. But the window looks as follows:
![argument completion](http://scrin.org/i/16/04/tmpfnovzd6e5us.png)

flying-sheep:
should be `inspect_request`

balta2ar:
I also occasionally see in the console this:
```
[1] "Got unhandled msg_type:" "comm_open"
```

poplav:
@balta2ar , Comm support was recently added #253.  It's likely not on the tag/version your using.

balta2ar:
Indeed, I've just noticed in the log: `[1] "Got unhandled msg_type:" "inspect_request"`

flying-sheep:
`inspect_request` is now handled, but still doesn’t do signatures

kmishra9:
Following — this would make Jupyter Notebooks truly competitive w/ R Notebooks in RStudio as the lack of quick documentation support is what is holding me back from using them all the time.

flying-sheep:
What do you propose? Currently, IRkernel finds the word under the cursor and inspects that one.

Would you say that it should instead find the innermost function call if possible? like:

```r
foo(one, two(two.one, two.two), three)
 ↑    ↑                 ↑
 |    |                 Shift-Tab here: inspects two, not two.two
 |    Shift-Tab here: inspects foo
 Shift-Tab here: also inspects foo

kmishra9:
Ah, so if Shift-Tab finally works for documentation support (I think the implementation proposed in the diagram sounds very reasonable too) I’ll have to give it a try. I’ll report back here if I can think of any suggestions/pain points to iterate on.

PR #441, opened 2017-01-26
Support `inspect_request` type message.

ngr-t:
Dear IRkernel maintainers,

In this PR I modified IRkernel to reply `inspect_request` type message, which is currently not supported. Therefore `shift+tab` key works now, to show a tooltip or page for object inspection as below.

tooltip

I intended this is just a draft proposal, there will be much to improve. Especially we should discuss about the format of object inspection.

## Motivation

Though there is already well-working pager to show help page within Jupyter Notebook, there are requests to get object inspection in a tooltip as #299. Also there are and will be some apps other than Jupyter Notebook which use Jupyter messaging protocol (like JupyterLab or Hydrogen package for Atom editor) and they would use `inspect_request` message. Therefore I think supporting it is so helpful.

## Format of object inspection

Below is a brief description of the format of object inspection in my current implementation.

### Sections of object inspection

Basically an object inspection has sections as

- "Class attribute"
- "Printed form"
- "Help document".

Some sections will be omitted if we cannot extract the information.

The value of `base::class()` is used for "Class attribute" section.

The value of `IRdisplay::prepare_mimebundle()` is used for the content of "Printed form".

"Help document" uses the value of `IRdisplay::prepare_mimebundle()`, but call it with putting `?` at the head of the inspected object name.

Maybe there are several things we should add to it.

### `data` types

`data` object in `inspect_reply` message has two type, which are `text/plain` and `text/html`, for now.

`IRdisplay::prepare_mimebundle()` returns also `text/latex` type and `text/markdown` type content, so supporting them should be not so difficult.

---

You can try it with installing this branch from my repo.
```{R}
> devtools::install_github("ngr-t/IRkernel", ref = "support_inspect_request")
```

I hope this PR interests you and helps IRkernel to support good object inspection.

Thanks!

### DIFF
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -244,11 +244,22 @@ complete = function(request) {
 },
 
 inspect = function(request) {
-    make_section_header_plain <- function (txt) paste0("# ", txt, ":\n")
-    make_section_header_html <- function (txt) paste0("<h1>", txt, ":</h1>\n")
     # 5.0 protocol:
     code <- request$content$code
     cursor_pos <- request$content$cursor_pos
+
+
+    section_templates <- list(
+        "text/plain" = "# %1s:\n%2s\n\n",
+        "text/html" = "<h1> %1s:</h1>\n%2s\n\n")
+
+    add_new_section <- function (data, section_name, new_data) {
+        for (mime in names(section_templates)) {
+            new_section <- sprintf(section_templates[[mime]], section_name, new_data[[mime]])
+            data[[mime]] <- paste0(data[[mime]], new_section)
+        }
+        return(data)
+    }
 
     # Get token under the cursor_pos.
     # There may be better ways to do that than this.
@@ -266,38 +277,15 @@ inspect = function(request) {
         data <- list()
 
         class_data <- tryCatch(
-            eval(parse(text = paste0("class(", token, ")"))),
-            error = function (e) character(0))
-        if (length(class_data) != 0) {
-            section_title <- "Class attribute"
-            body <- paste(class_data, collapse = " ")
-            data[["text/plain"]] <- paste0(
-                make_section_header_plain(section_title),
-                body,
-                "\n\n")
-            data[["text/html"]] <- paste0(
-                make_section_header_html(section_title),
-                body,
-                "\n")
-        }
+            IRdisplay::prepare_mimebundle(class(token)),
+            error = function (e) list())
+        data <- add_new_section(data, "Class attribute", class_data)
 
         code_to_get_print_mimebundle <- paste0("IRdisplay::prepare_mimebundle(", token, ")")
         print_data <- tryCatch(
             eval(parse(text = code_to_get_print_mimebundle))$data,
             error = function (e) list())
-        if (length(print_data) != 0) {
-            section_title <- "Printed form"
-            data[["text/plain"]] <- paste0(
-                data[["text/plain"]],
-                make_section_header_plain(section_title),
-                print_data[["text/plain"]],
-                "\n\n")
-            data[["text/html"]] <- paste0(
-                data[["text/html"]],
-                make_section_header_html(section_title),
-                print_data[["text/html"]],
-                "\n")
-        }
+        data <- add_new_section(data, "Printed form", print_data)
 
         code_to_get_help_mimebundle <- paste0("IRdisplay::prepare_mimebundle(?", token, ")")
         help_data <- tryCatch({
@@ -306,19 +294,7 @@ inspect = function(request) {
                 help_data <- eval(parse(text = code_to_get_help_mimebundle))$data
             }},
             error = function (e) list())
-        if (length(help_data) != 0) {
-            section_title <- "Help document"
-            data[["text/plain"]] <- paste0(
-                data[["text/plain"]],
-                make_section_header_plain(section_title),
-                help_data[["text/plain"]],
-                "\n\n")
-            data[["text/html"]] <- paste0(
-                data[["text/html"]],
-                make_section_header_html(section_title),
-                help_data[["text/html"]],
-                "\n")
-        }
+        data <- add_new_section(data, "Help document", help_data)
         found <- (length(data) != 0)
     }
     send_response('inspect_reply', request, 'shell', list(
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -278,7 +278,7 @@ inspect = function(request) {
         tryCatch(
             {
                 # In many cases `get(token)` works, but it does not
-                # in the cases such as `token` is a numeric constants or reserved words.
+                # in the cases such as `token` is a numeric constant or reserved word.
                 # Therefore `eval()` is used here.
                 obj <- eval(parse(text = token))
 
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -248,7 +248,6 @@ inspect = function(request) {
     code <- request$content$code
     cursor_pos <- request$content$cursor_pos
 
-
     section_templates <- list(
         'text/plain' = '# %1s:\n%2s\n\n',
         'text/html' = '<h1> %1s:</h1>\n%2s\n\n')
@@ -265,9 +264,9 @@ inspect = function(request) {
     # There may be better ways to do that than this.
     token <- ''
     for (i in seq(cursor_pos, nchar(code))) {
-        topic_candidate <- utils:::.guessTokenFromLine(code, i)
-        if (!grepl(token, topic_candidate)) break
-        token <- topic_candidate
+        token_candidate <- utils:::.guessTokenFromLine(code, i)
+        if (nchar(token_candidate) == 0) break
+        token <- token_candidate
     }
 
     if (nchar(token) == 0) {
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -250,8 +250,8 @@ inspect = function(request) {
 
 
     section_templates <- list(
-        "text/plain" = "# %1s:\n%2s\n\n",
-        "text/html" = "<h1> %1s:</h1>\n%2s\n\n")
+        'text/plain' = '# %1s:\n%2s\n\n',
+        'text/html' = '<h1> %1s:</h1>\n%2s\n\n')
 
     add_new_section <- function (data, section_name, new_data) {
         for (mime in names(section_templates)) {
@@ -263,7 +263,7 @@ inspect = function(request) {
 
     # Get token under the cursor_pos.
     # There may be better ways to do that than this.
-    token <- ""
+    token <- ''
     for (i in seq(cursor_pos, nchar(code))) {
         topic_candidate <- utils:::.guessTokenFromLine(code, i)
         if (!grepl(token, topic_candidate)) break
@@ -279,22 +279,22 @@ inspect = function(request) {
         class_data <- tryCatch(
             IRdisplay::prepare_mimebundle(class(token)),
             error = function (e) list())
-        data <- add_new_section(data, "Class attribute", class_data)
+        data <- add_new_section(data, 'Class attribute', class_data)
 
-        code_to_get_print_mimebundle <- paste0("IRdisplay::prepare_mimebundle(", token, ")")
+        code_to_get_print_mimebundle <- paste0('IRdisplay::prepare_mimebundle(', token, ')')
         print_data <- tryCatch(
             eval(parse(text = code_to_get_print_mimebundle))$data,
             error = function (e) list())
-        data <- add_new_section(data, "Printed form", print_data)
+        data <- add_new_section(data, 'Printed form', print_data)
 
-        code_to_get_help_mimebundle <- paste0("IRdisplay::prepare_mimebundle(?", token, ")")
+        code_to_get_help_mimebundle <- paste0('IRdisplay::prepare_mimebundle(?', token, ')')
         help_data <- tryCatch({
-            help_filename <- as.character(eval(parse(text = paste0("?", token))))
+            help_filename <- as.character(eval(parse(text = paste0('?', token))))
             if (length(help_filename) > 0) {
                 help_data <- eval(parse(text = code_to_get_help_mimebundle))$data
             }},
             error = function (e) list())
-        data <- add_new_section(data, "Help document", help_data)
+        data <- add_new_section(data, 'Help document', help_data)
         found <- (length(data) != 0)
     }
     send_response('inspect_reply', request, 'shell', list(
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -121,6 +121,7 @@ handle_shell = function() {
         history_request     = history(msg),
         complete_request    = complete(msg),
         is_complete_request = is_complete(msg),
+        inspect_request     = inspect(msg),
         shutdown_request    = shutdown(msg),
         print(c('Got unhandled msg_type:', msg$header$msg_type)))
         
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks! This looks good to me at a quick glance; I'll let others review the details.

ngr-t:
Thank you for reply.

I've come up with the idea that printing attributes error be more helpful.

R/kernel.r
```diff
@@ -242,6 +243,92 @@ complete = function(request) {
         cursor_end = start_position + nchar(c.info$token)))
 },
 
+inspect = function(request) {
+    make_section_header_plain <- function (txt) paste0("# ", txt, ":\n")
```
flying-sheep: style: we use single quotes

R/kernel.r
```diff
@@ -242,6 +243,92 @@ complete = function(request) {
...
+    make_section_header_html <- function (txt) paste0("<h1>", txt, ":</h1>\n")
+    # 5.0 protocol:
+    code <- request$content$code
+    cursor_pos <- request$content$cursor_pos
+
+    # Get token under the cursor_pos.
+    # There may be better ways to do that than this.
+    token <- ""
+    for (i in seq(cursor_pos, nchar(code))) {
```
flying-sheep: why a loop? isn’t `cursor_pos` either 1 to the left or exactly on the first character of the token (if a token is below the cursor at all?)

this way you’ll find a token later in the line if the cursor is above e.g. punctuation.

R/kernel.r
```diff
@@ -242,6 +243,92 @@ complete = function(request) {
...
+        print_data <- tryCatch(
+            eval(parse(text = code_to_get_print_mimebundle))$data,
+            error = function (e) list())
+        if (length(print_data) != 0) {
+            section_title <- "Printed form"
+            data[["text/plain"]] <- paste0(
+                data[["text/plain"]],
+                make_section_header_plain(section_title),
+                print_data[["text/plain"]],
+                "\n\n")
```
flying-sheep: i’d rather do:

```r
section_templates <- list(
	'text/plain' = "%s# %s:\n%s\n\n",
	'text/html' = "%s<h1>%s:</h1>\n%s\n\n")
```

at the beginning and then here:

```r
for (mime in names(section_templates)) {
	data[[mime]] <- sprintf(section_templates[[mime]], data[[mime]], section_title, print_data[[mime]])
}
```

R/kernel.r
```diff
@@ -242,6 +243,92 @@ complete = function(request) {
...
+    }
+
+    if (nchar(token) == 0) {
+        data <- list()
+        found <- FALSE
+    } else {
+        data <- list()
+
+        class_data <- tryCatch(
+            eval(parse(text = paste0("class(", token, ")"))),
```
flying-sheep: all evals from this point on are unnecessary.

just `obj <- get(token)` should get you the object, then you can just call functions normally:

`class(obj)` works, `obj_help <- help(obj)` does as well, and then you can call `IRdisplay::prepare_mimebundle` on both `obj` and `obj_help` directly.

R/kernel.r
```diff
@@ -242,6 +243,92 @@ complete = function(request) {
...
+                "\n\n")
+            data[["text/html"]] <- paste0(
+                data[["text/html"]],
+                make_section_header_html(section_title),
+                print_data[["text/html"]],
+                "\n")
+        }
+
+        code_to_get_help_mimebundle <- paste0("IRdisplay::prepare_mimebundle(?", token, ")")
+        help_data <- tryCatch({
```
flying-sheep: `help_data` and `print_data` are created using almost identical code.

if there’s still more than 3 lines left after applying all my other ideas, make a function.

ngr-t:
@flying-sheep Thank you for reviewing. I begin to fix!

> are there objects that have both help and a repr?

Datasets included by packages have both. We can also get repr for functions, though in many cases we want to see help more than implementation (maybe setting `detail_level` is better choice about function.)

> and/or will that occur if you have e.g. a variable in your workspace shadowing a function with a help page?

Sorry, I didn't care about this case. We get help document if we call `help` or `?` for objects with help document, even though it's masked.

flying-sheep:
Hmm, i’m not even sure if masking should change things without giving additional info. So the ideal version would be to clearly say something like:

> Help topic for function `toolz::foo`:
> 
> …
> 
> Content of the `foo` object in the workspace:
> 
> …

But you don’t need to bother about this now. I think we can just as well add that later.

R/kernel.r
ngr-t: Yes, `class(token)` or `IRdisplay::prepare_mimebundle(obj)` works. There is no good reason to use `eval()`.
Also `IRdisplay::prepare_mimebundle(help(token))` basically work, but we can't get help documents in the case `token` is `pkg::topic` or `pkg:::topic` form. We need to split it into `pkg` and `topic` manually and I felt it a little bit hassle than using `eval()`. But I'll try to make it simple without `eval()`.

R/kernel.r
ngr-t: The reason for the loop is that `tools:::.guessTokenFromLine()` checks only characters before `cursor_pos`. It is necessary to check characters after `cursor_pos` in some way if we want to get the correct token on invoking `inspect_request` when the cursor is within a token.
Some combination of string manipulate functions may do the trick, but I thought the loop is easier to understand.

But I must admit that my implementation causes a problem when the cursor is above punctuations, white spaces or so. I should finish the loop if the first call of function returns an empty string. Because `tools:::.guessTokenFromLine()` returns an empty string if `cursor_pos` is on the characters not used in tokens.

Or, is there a tokenizer function which fits this purpose more? If not, I think of putting the code to finish the loop on the condition I mentioned above. How do you think?

R/kernel.r
ngr-t: I found that `get(token)` fails when the content of `token` is a built-in constants such as `1` or `NULL`. We can also get repr or helps for them, so the object inspection for them should work.

ngr-t:
There are so many corner cases to deal with that I'd better to add test code for them. It seems that you've written the test codes about messaging in `IRkernelTests` class in `tests/testthat/test_ir.py` and I should put my test codes here,  is that OK?

takluyver:
There are two parts to the tests - the R files test functions in the code, as is normal for an R project. For details of what's going on, it's more efficient to pull out pieces into their own functions and test them from R. The Python file is used for tests which start up a kernel and talk to it, so it checks that the whole system is working together. It would be good to add at least one test of using an inspect_request to the Python code, to make sure that it's hooked up correctly.

takluyver:
N.B. tests are failing on Travis. It was failing before you added tests as well - I don't know if something in this PR broke it or it was already broken, but could you have a look?

R/kernel.r
ngr-t: This implementation is very simple and easier to extend, but I caught an error when an object has long help document due to the limitation that `sprintf()` can't deal a string longer than 8192 bytes. It seems that using `paste()` is safer from this point of view.
Maybe your point is to avoid repetitions in codes which have a potential risk to cause inconsistency in format, isn't? If so, how about a thing like this?
```R
plain_text_content_builder <- function (content, section_title, new_section) paste0(content, '\n\n#: ', section_title,...)
html_content_builder <- function (...) ...

content_builders <- list(
    'text/plain' = plain_text_content_builder,
    'text/html' = html_content_builder)
```
and
```R
for (mime in names(content_builders)) {
    data[[mime]] <- content_builders[[mime]](data[[mime]], section_title, print_data[[mime]])
}
```

ngr-t:
@takluyver 
I found that the test below fails. (I didn't touch here)
```Python
    def test_irkernel_df_default_rich_output(self):
        """data.frame rich representation"""
        code = 'data.frame(x = 1:3)'
        reply, output_msgs = self._execute_code(code)
        
        # we currently send three formats: text/plain, html, and latex
        data = output_msgs[0]['content']['data']
        self.assertEqual(len(data), 3, data.keys())
```

I fixed the last lines as below (just because error message says the length is 4)

```
        self.assertEqual(len(data), 4, data.keys())
```

then the test passes in my environment. I'm not sure about the reason.

takluyver:
Hmm. The error message should show the new list of keys - what does  it say? It might be a legitimate addition that requires changing the tests, or it might be something spurious.

ngr-t:
@takluyver 

Below is the message I got. I run test at the commit [ce25ae2](https://github.com/IRkernel/IRkernel/pull/441/commits/ce25ae2607637e0712b3885b7e222bc12d185bde)

```R

> devtools::test()
Loading IRkernel
Loading required package: testthat
Testing IRkernel
default options: ......
kernel: [InstallKernelSpec] Installed kernelspec testir in /home/ngr-t/.local/share/jupyter/kernels/testir
..S.............SSS.S..S1..........S....2[RemoveKernelSpec] Removed /home/ngr-t/.local/share/jupyter/kernels/testir
.
utils: ...

Skipped ------------------------------------------------------------------------
1. test ir.IRkernelTests.test_clear_output - <no description>

2. test ir.IRkernelTests.test_error - <no description>

3. test ir.IRkernelTests.test_execute_result - <no description>

4. test ir.IRkernelTests.test_execute_stderr - <no description>

5. test ir.IRkernelTests.test_history - <no description>

6. test ir.IRkernelTests.test_inspect - <no description>

7. test ir.IRkernelTests.test_pager - <no description>

Failed -------------------------------------------------------------------------
1. Failure: data.frame rich representation -------------------------------------
Traceback (most recent call last):
  File "/home/ngr-t/Documents/GitHub/IRkernel/tests/testthat/test_ir.py", line 169, in test_irkernel_df_default_rich_output
    self.assertEqual(len(data), 3, data.keys())
AssertionError: 4 != 3 : dict_keys(['text/html', 'text/markdown', 'text/latex', 'text/plain'])



2. Failure: kernel tests pass --------------------------------------------------
close(con) not equal to 0.
1/1 mismatches
[1] 256 - 0 == 256


DONE ===========================================================================
```

takluyver:
Thanks, so Markdown is what has been added and is making the tests fail. @flying-sheep, I take it that's a deliberate addition? If so, @ngr-t can fix the test in this PR. :-)

ngr-t:
I added tests and fixed the codes into pass all the tests I wrote (I didn't fix the test mentioned in the adjacent comments in the commits).
Sorry about I'm not yet fully following the changes requested by @flying-sheep , but I think some of them are still arguable. Anyway it should be much cleaner than the first version I submit, thanks.

I've left comments to each change request if there remains a point to discuss about.

R/kernel.r
```diff
@@ -242,6 +243,75 @@ complete = function(request) {
...
+        if (is.null(new_content)) return(content)
+        title <- paste0('# ', section_name, ':\n')
+        return(paste0(content, title, new_content, '\n', sep = '\n'))
+    }
+
+    add_new_section_html <- function (content, section_name, new_content) {
+        if (is.null(new_content)) return(content)
+        title <- paste0('<h1>', section_name, ':</h1>\n')
+        return(paste0(content, title, new_content, '\n', sep = '\n'))
+    }
```
flying-sheep: duplicated code again. the two function only differ by their second line, so just do

```r
create_title <- list(
	'text/plain' = function(section_name) paste0('# ', section_name, ':\n'),
	'text/html'  = function(section_name) paste0('<h1>', section_name, ':</h1>\n'))
```

and put the rest of the code into `add_new_section`.

flying-sheep:
hi, i fixed that test on master. if you rebase it should be green.

can you explain what you mean by the following?

> Avoid using `sprintf()` on constructing the messages due to errors caused by its limitation about character length.

R/kernel.r
```diff
@@ -242,6 +243,75 @@ complete = function(request) {
...
+        token_candidate <- utils:::.guessTokenFromLine(code, i)
+        if (nchar(token_candidate) == 0) break
+        token <- token_candidate
+    }
+
+    data <- namedlist()
+    if (nchar(token) != 0) {
+        tryCatch(
+            {
+                obj <- eval(parse(text = token))
```
flying-sheep: please add a comment why you don’t use “get(as.character(token))” here (if there’s no reason, please use that instead)

R/kernel.r
```diff
@@ -242,6 +243,75 @@ complete = function(request) {
...
+                class_data <- IRdisplay::prepare_mimebundle(class(obj))$data
+                data <- add_new_section(data, 'Class attribute', class_data)
+
+                print_data <- IRdisplay::prepare_mimebundle(obj)$data
+                data <- add_new_section(data, 'Printed form', print_data)
+            },
+            error = function (e) NULL)
+        tryCatch(
+            {
+                help_obj <- eval(parse(text = paste0('?', token)))
```
flying-sheep: please also add a comment why you don’t use `help_obj <- help(obj)` here.

i assume it is because there might be no object and still a help topic, am i right?

tests/testthat/test_ir.py
```diff
@@ -239,6 +242,64 @@ def test_should_not_increment_history(self):
...
+        test_token_is_ok('c')
+        # Function name with namespace
+        test_token_is_ok('base::c')
+        # Function not exported from namespace
+        test_token_is_ok('tools:::.Rd2pdf')
+        # User-defined variable
+        test_token_is_ok(
+            'x',
+            preprocess='x <- 1',
+            postprocess='rm("x")')
```
flying-sheep: python style: use trailing commas and the closing brace on its own line please:

```py
foo(
    'a',
    b='c',
)
```

R/kernel.r
flying-sheep: OK!

R/kernel.r
flying-sheep: see the other review below 😄

R/kernel.r
flying-sheep: i couldn’t find one. basically we’d need to adapt the `.guessTokenFromLine` code ourselves or so…

R/kernel.r
ngr-t: For example,

```R
token1 <- 'seq'
help(token)
```

returns help doc for `seq()`, but 

```R
token2 <- 'base::seq'
help(token)
```

does not, at least in my environment (R 3.3.2).
On the other hand,

```R
 eval(parse(text = paste0("?", token1)))
 eval(parse(text = paste0("?", token2)))
```

both works. 

We can split `token2` into `topic` and `pkg` by colons manually  and call `help(topic, package = pkg)`. We have to care about the cases for both 2 colons (exported object) and 3 colons (not exported) if we do this. I feel it lengthy and unnecessarily complex, but I just can't figure out which is a preferred choice for this project.

R/kernel.r
ngr-t: For the cases such as `token` is `"1"` or `"NULL"`. We can't get the objects for them by `get(token)`, but we can by `eval(parse(text = token))`.
Given above, I think of continuing to use `eval()` and adding comments for the lines. Is it OK?

ngr-t: Sorry, I overlooked your reply on the review for here. I'll add comments here.

tests/testthat/test_ir.py
ngr-t: OK, I follow.

R/kernel.r
ngr-t: OK, I agree it's simpler. I'll do that.

R/kernel.r
flying-sheep: OK, let’s merge it like this for now. i’ll try my hand at crafting a regex that finds token and (optional) package from the code.

this would replace the loop above, and not be worse than `.guessTokenFromLine` (since that one also uses a regex)

R/kernel.r
```diff
@@ -242,6 +243,71 @@ complete = function(request) {
...
+        'text/plain' = '# %s:\n',
+        'text/html' = '<h1>%s:</h1>\n')
+    # Function to add a section to content.
+    add_new_section <- function (data, section_name, new_data) {
+        for (mime in names(new_section_adders)) {
+            new_content <- new_data[[mime]]
+            if (is.null(new_content)) next
+            title <- sprintf(title_templates, section_name)
+            # use paste0 since sprintf cannot deal with format strings > 8192 bytes
+            data[[mime]] <- paste0(content, title, new_content, '\n', sep = '\n')
```
flying-sheep: where does `content` come from? isn’t this `data[[mime]]`?

R/kernel.r
```diff
@@ -242,6 +243,71 @@ complete = function(request) {
...
+inspect = function(request) {
+    # 5.0 protocol:
+    code <- request$content$code
+    cursor_pos <- request$content$cursor_pos
+
+    title_templates <- list(
+        'text/plain' = '# %s:\n',
+        'text/html' = '<h1>%s:</h1>\n')
+    # Function to add a section to content.
+    add_new_section <- function (data, section_name, new_data) {
```
flying-sheep: style `function(`

R/kernel.r
```diff
@@ -242,6 +243,71 @@ complete = function(request) {
...
+                # Therefore `eval()` is used here.
+                obj <- eval(parse(text = token))
+
+                class_data <- IRdisplay::prepare_mimebundle(class(obj))$data
+                data <- add_new_section(data, 'Class attribute', class_data)
+
+                print_data <- IRdisplay::prepare_mimebundle(obj)$data
+                data <- add_new_section(data, 'Printed form', print_data)
+            },
+            error = function (e) NULL)
```
flying-sheep: style `function(` or just use `identity` or so

R/kernel.r
```diff
@@ -242,6 +243,71 @@ complete = function(request) {
...
+        tryCatch(
+            {
+                # `help(token)` is  not used here because it does not works
+                # in the cases `token` is in `pkg::topic`or `pkg:::topic` form.
+                help_obj <- eval(parse(text = paste0('?', token)))
+                help_data <- IRdisplay::prepare_mimebundle(help_obj)$data
+                data <- add_new_section(data, 'Help document', help_data)
+            },
+            error = function (e) NULL)
```
flying-sheep: the same

R/kernel.r
```diff
@@ -242,6 +243,71 @@ complete = function(request) {
...
+    }
+    found <- (length(data) != 0)
+    send_response('inspect_reply', request, 'shell', list(
+        status = 'ok',
+        found = found,
+        data = data,
+        metadata = namedlist()
+        ))
```
flying-sheep: style: put `))` on previous line

R/kernel.r
```diff
@@ -242,6 +243,71 @@ complete = function(request) {
...
+            {
+                # `help(token)` is  not used here because it does not works
+                # in the cases `token` is in `pkg::topic`or `pkg:::topic` form.
+                help_obj <- eval(parse(text = paste0('?', token)))
+                help_data <- IRdisplay::prepare_mimebundle(help_obj)$data
+                data <- add_new_section(data, 'Help document', help_data)
+            },
+            error = function (e) NULL)
+    }
+    found <- (length(data) != 0)
```
flying-sheep: style: no `()` around the expression

R/kernel.r
ngr-t: Oh, so sorry for silly bugs. I had a mistake in refactoring.

ngr-t:
Fixed requested points about styles and bugs. All the tests passed, this should be OK.

flying-sheep:
great stuff, thank you very much!

ngr-t:
Thank you, too!

takluyver:
Awesome, thanks @ngr-t - this was quite substantial for a first contribution to the project. :-)

Issue #445, opened 2017-02-17
error in should_store_history


phil8192:
i get this when executing cell from within [emacs ipython notebook](https://github.com/millejoh/emacs-ipython-notebook):

```r
Error in !is_silent() && should_store_history() : 
  invalid 'y' type in 'x && y'
Calls: <Anonymous> -> <Anonymous> -> handle_shell -> <Anonymous>
Execution halted
```

PR #446, opened 2017-02-17
fix for IRkernel/IRkernel#445

phil8192:


flying-sheep:
i’d rather do:

```r
sh <- current_request$content$store_history
!is.null(sh) && sh
```

takluyver:
Thanks @phil8192 :-)

PR #449, opened 2017-03-03
Add dependency 'uuid' in installation process

XD-DENG:
I was reminded that package 'uuid' is not installed when I try to install IRkernel after I have run `install.packages(c('repr', 'IRdisplay', 'crayon', 'pbdZMQ', 'devtools'))`.

It may be good to add it in the command

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ For detailed requirements and install instructions see [irkernel.github.io](http
 We will soon submit the IRkernel package to CRAN. Until then, you can install it via the `devtools` package:
 
 ```R
-install.packages(c('repr', 'IRdisplay', 'crayon', 'pbdZMQ', 'devtools'))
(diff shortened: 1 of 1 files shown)


flying-sheep:
actually idk why we have that line at all.

```r
install.packages('devtools')
devtools::install_github('IRkernel/IRkernel')
```

should be sufficient. what is the error message? does this work?

```r
devtools::install_github('IRkernel/IRkernel', depenencies = NA)
```

XD-DENG:
Was simply warned that dependency `uuid` is not installed when I tried to install the package. 

I guess possibly because I tried to install it locally from the **.zip** file (as `install_github()` is not working due to the firewall setting on my another PC). 

If that's the case, feel free to close this issue.

flying-sheep:
yes, that’s it. if you had used `devtools::install` or so, it’d have resolved all dependencies for you

Issue #429, opened 2016-12-01
Ask for password prompt


nfx:
How would it be possible to write ask-for-password prompt within this kernel's notebook, so that password chars would not be displayed?

takluyver:
We'd need to implement the [input_request message](http://jupyter-client.readthedocs.io/en/latest/messaging.html#messages-on-the-stdin-router-dealer-channel) from the Jupyter spec, which is issue #199. Once that's done, providing a way to use that with hidden input is simple.

jgbradley1:
I'd like to point out that the [getPass](https://github.com/wrathematics/getPass) package may provide this capability. It will depend on how [#199](https://github.com/IRkernel/IRkernel/issues/199) is solved.

jgbradley1:
This issue should be closed now.

flying-sheep:
thanks!

Issue #236, opened 2015-12-19
identify as interactive


flying-sheep:
- [ ] `options(menu.graphics = FALSE)` (#235)
- [ ] implement stdin protocol (#199)
- [ ] (maybe fits) interactive rgl session (#200)

@takluyver i’d like to add labels everywhere – could you upgrade my organization permissions please?

janschulz:
At least stdin and input_request should be implemented for 0.7

flying-sheep:
OK, this is pretty hard:

1. when passing `--interactive` to force the internal `R_interactive` variable to be TRUE, R crashes. No idea why, as there’s no error.
2. the R-level [`interactive()`](https://github.com/wch/r-source/blob/e5b21d0397c607883ff25cca379687b86933d730/src/main/sysutils.c#L223) function just checks that variable
3. they say [they support 4 methods of input](https://github.com/wch/r-source/blob/e5b21d0397c607883ff25cca379687b86933d730/src/gnuwin32/system.c#L166-L170) none of which resembles ours.

we might need to work with upstream R on this one.

flying-sheep:
created a wishlist entry: https://bugs.r-project.org/bugzilla/show_bug.cgi?id=17134

Issue #199, opened 2015-10-06
Implement stdin protocol


takluyver:
The R [readline](https://stat.ethz.ch/R-manual/R-devel/library/base/html/readline.html) function should, inside IRkernel, use the [Jupyter protocol stdin messages](http://jupyter-client.readthedocs.org/en/latest/messaging.html#messages-on-the-stdin-router-dealer-sockets) to prompt the user for input.

Slight complication: should we then make [interactive()](https://stat.ethz.ch/R-manual/R-devel/library/base/html/interactive.html) return TRUE? That appears to be the signal that you can use readline(), but it also affects some other bits of behaviour, which we may not want.

flying-sheep:
there is AFAIK only `readline` and `menu` directly using the terminal, everything else uses `stdin()`

pieterlukasse:
+1 for  The R readline function should, inside IRkernel, use the Jupyter protocol stdin messages to prompt the user for input.

forus:
it would be nice to have this feature. +1

flying-sheep:
@pieterlukasse @forus sure it would, but i have no idea how to do this cleanly. if you want this, please consider researching how to best intercept the input methods for R’s readline and menu functions.

i.e. if there’s a better way than

```r
unlockBinding('readline', baseenv())
assign('readline', function(prompt = '') { ... }, envir = baseenv())
lockBinding('readline', baseenv())
```

nfx:
+1

PR #452, opened 2017-03-21
Implemented user input for readline and password functionality

jgbradley1:
This addresses issues #199 and #429. This may also address some of issues #85 and #236.

### DIFF
--- a/R/execution.r
+++ b/R/execution.r
@@ -56,12 +56,12 @@ format_stack <- function(calls) {
     paste0(tb, line_refs)
 }
 
-
 #' @importFrom utils capture.output
 Executor <- setRefClass(
     Class = 'Executor',
     fields = list(
         send_response         = 'function',
+        handle_stdin          = 'function',
         abort_queued_messages = 'function',
         execution_count       = 'integer',
         payload               = 'list',
@@ -118,6 +118,28 @@ quit = function(save = 'default', status = 0, runLast = TRUE) {
     }
     if (save) NULL  # TODO: actually save history
     payload <<- c(.self$payload, list(list(source = 'ask_exit', keepkernel = FALSE)))
+},
+
+# noninteractive
+readline = function(prompt='') {
+    log_debug("entering custom readline")
+    send_response('input_request', current_request, 'stdin',
+        list(prompt = prompt, password = FALSE))
+    # wait for 'input_reply' response message
+    input = handle_stdin()
+    log_debug("exiting custom readline")
+    return(input)
+},
+
+# noninteractive 5.0 protocol:
+readpass = function(prompt='') {
+    log_debug("entering custom readpass")
+    send_response('input_request', current_request, 'stdin',
+        list(prompt = prompt, password = TRUE))
+    # wait for 'input_reply' response message
+    input = handle_stdin()
+    log_debug("exiting custom readpass")
+    return(input)
 },
 
 handle_error = function(e) tryCatch({
@@ -216,13 +238,19 @@ execute = function(request) {
     send_response('execute_input', request, 'iopub', list(
         code = request$content$code,
         execution_count = execution_count))
-        
+
     # Make the current request available to other functions
     current_request <<- request
     # reset ...
     payload <<- list()
     err <<- list()
     
+    # shade base::readline
+    add_to_user_searchpath('readline', .self$readline)
+    
+    # shade 'readpass' if it appears in other packages
+    add_to_user_searchpath('readpass', .self$readpass)
+
     # shade base::quit
     add_to_user_searchpath('quit', .self$quit)
     add_to_user_searchpath('q', .self$quit)
@@ -257,7 +285,7 @@ execute = function(request) {
     log_debug('Executing code: %s', request$content$code)
     
     warn_unicode_on_windows(request$content$code, .self$send_error_msg)
-    
+
     tryCatch(
         evaluate(
             request$content$code,
@@ -266,7 +294,7 @@ execute = function(request) {
             stop_on_error = 1L),
         interrupt = function(cond) interrupted <<- TRUE,
         error = .self$handle_error) # evaluate does not catch errors in parsing
-    
+
     if (!is_silent() && !is.null(last_recorded_plot)) {
         send_plot(last_recorded_plot)
     }
--- a/R/execution.r
+++ b/R/execution.r
@@ -56,6 +56,7 @@ format_stack <- function(calls) {
     paste0(tb, line_refs)
 }
 
+
 #' @importFrom utils capture.output
 Executor <- setRefClass(
     Class = 'Executor',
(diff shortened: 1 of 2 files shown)


R/execution.r
```diff
@@ -120,6 +120,28 @@ quit = function(save = 'default', status = 0, runLast = TRUE) {
     payload <<- c(.self$payload, list(list(source = 'ask_exit', keepkernel = FALSE)))
 },
 
+# noninteractive
+readline = function(prompt='') {
+    log_debug("entering custom readline")
```
flying-sheep: code style (here and below); please use:

* spaces around opearators and default argument `=`
* single quotes, except in strings containing `'` or method docstrings
* `<-` for assignment
* no useless `return`, i.e. only use `return` for early exit, never in terminal position

R/execution.r
```diff
@@ -216,13 +238,19 @@ execute = function(request) {
     send_response('execute_input', request, 'iopub', list(
         code = request$content$code,
         execution_count = execution_count))
-        
+
```
flying-sheep: please configure your editor so it doesn’t touch lines you didn’t touch. this whitespace (and the whitespace in multiple more lines below) should stay.

R/kernel.r
```diff
@@ -123,7 +123,7 @@ handle_shell = function() {
         is_complete_request = is_complete(msg),
         inspect_request     = inspect(msg),
         shutdown_request    = shutdown(msg),
-        print(c('Got unhandled msg_type:', msg$header$msg_type)))
+        log_debug(c('Got unhandled msg_type:', msg$header$msg_type)))
```
flying-sheep: good catch, thanks!

R/kernel.r
```diff
@@ -166,6 +166,27 @@ abort_queued_messages = function() {
...
+                 c(.pbd_env$ZMQ.PO$POLLIN)) # type
+
+        if (bitwAnd(zmq.poll.get.revents(1), .pbd_env$ZMQ.PO$POLLIN)) {
+            log_debug('stdin loop: found msg')
+            parts <- zmq.recv.multipart(sockets$stdin, unserialize = FALSE)
+            msg <- wire_to_msg(parts)
+            return(msg$content$value)
+        } else {
+            # else shouldn't be possible
+            log_debug('stdin loop: zmq.poll returned but no message found?')
```
flying-sheep: this should be `log_error`. i’d also prefer if you do the following instead of “early” exit by `return`

```r
if (...) {
    ...
    msg$content$value
} else {
    # ...
    log_error(...)
    NULL
}

File: R/kernel.r [hunk 5]
```diff
@@ -372,7 +393,8 @@ initialize = function(connection_file) {
     zmq.bind(sockets$shell,   url_with_port('shell_port'))
 
     executor <<- Executor$new(send_response = .self$send_response,
-        abort_queued_messages = .self$abort_queued_messages)
+                              handle_stdin = .self$handle_stdin,
```
flying-sheep: please rather do:

```r
executor <<- Executor$new(
    send_response         = .self$send_response,
    handle_stdin          = .self$handle_stdin,
    abort_queued_messages = .self$abort_queued_messages)
```

jgbradley1:
Ok I made all the style changes. It looks like whitespace is back in where it should be. Let me know if there is anything else to address.

flying-sheep:
looks great. i’ll just test and then merge

flying-sheep:
oh, i was stupid! i didn’t think of the while loop when asking you to change the code to not use return. inside of a while loop, `return` is obviously the right thing!

thanks!

jgbradley1:
I updated my fork and shaded ```interactive```. It works fine and nothing weird happens but I don't know what tests need to be run to really test out the effects.

flying-sheep:
i think we should grep the places where `interactive` is used in the default R packages (`base` and so on), and test if those work.

jgbradley1:
What I've found so far. Shading interactive does not seem like it will be as straightforward as readline. After shading interactive to return TRUE, testing out menu() still errors out because interactive returns FALSE

Forcing R to run in interactive mode via `--interactive` in the kernelspec still causes the kernel to crash as you reported in #236.

It may take some work but an alternative option is to run R in a way to continuously read from stdin. A simple example would look something like

```r
#!/usr/local/bin/Rscript

OpenRead <- function(arg) {
   if (arg %in% c("-", "/dev/stdin")) {
      file("stdin", open = "r")
   } else if (grepl("^/dev/fd/", arg)) {
      fifo(arg, open = "r")
   } else {
      file(arg, open = "r")
   }
}

# read from stdin in a while loop
input <- OpenRead('-')

while(TRUE) {
  result <- tryCatch({
    readLines(input, n=1)
    }, warning = function(w) {
      NULL
    }, error = function(e) {
      NULL
    })
  if (result == '')
    break
  if (!is.null(result))
    eval(parse(text = result))
}
```

flying-sheep:
ah yes. the same thing is true for `readline()`: when packages (instead of the user) call one of those, they actually call a specific version, not the one found in the user searchpath. so we might need to replace them *in the package* (via `unlockBinding`)

here‘s where `readline` is called. i bet those also won’t work with our current code:

```r
rg 'readline\(' R-3.2.1/src/library/**/*.r
R-3.2.1/src/library/base/R/library.R
126:                readline("press RETURN to view license")

R-3.2.1/src/library/base/R/New-Internal.R
173:readline <- function(prompt = "") .Internal(readline(prompt))

R-3.2.1/src/library/utils/R/demo.R
118:        readline("\nType  <Return>   to start : ")

R-3.2.1/src/library/utils/R/help.request.R
33:     readline(paste(paste(strwrap(paste(...)), collapse="\n"),
69:    post <- readline("Have you read the posting guide? (y/n) ")
71:    FAQ <- readline("Have you checked the FAQ? (y/n) ")
73:    intro <- readline("Have you checked An Introduction to R? (y/n) ")
78:    rsitesearch <- readline("Have you looked on RSiteSearch? (y/n) ")
123:    change <- readline(paste("Would you like to change your subject line:",
127:    subject <- readline("Enter subject: \n")

R-3.2.1/src/library/utils/R/packages2.R
275:                ans <- readline(paste(sprintf(msg, userdir), " (y/n) "))
418:                    res <- readline("y/n: ")
440:                    res <- readline("y/n: ")

R-3.2.1/src/library/utils/R/packages.R
317:            answer <- substr(readline("Update (y/N/c)?  "), 1L, 1L)

R-3.2.1/src/library/utils/R/packageStatus.R
169:            answer <- substr(readline("Update (y/N/x)?  "), 1L, 1L)

R-3.2.1/src/library/utils/R/unix/create.post.R
89:        answer <- readline(paste0("Email the ", description, " now? (yes/no) "))
```

jgbradley1:
You're right. I was able to verify by running

```r
x <- packageStatus()
upgrade(x)
```
You need an outdated package to test with the above code. You'll notice the normal prompt to install or not doesn't wait for user input. Using your suggestion for `unlockBinding` I came up with a solution though. Unlock the binding for base function calls and replace them. This would apply to `quit` as well. I was able to test the above code on my current branch and it works. I'll make a pull request if you're satisfied with it.

As for shading `interactive` by with new approach, it still does not work, easily tested by
```r
menu(c('A','B','C'))
```

PR #457, opened 2017-03-30
Improve the shading of readline and quit

jgbradley1:
This fix better shades `readline` and `quit` to work when they are called inside other packages.

### DIFF
--- a/R/environment_shadow.r
+++ b/R/environment_shadow.r
@@ -6,6 +6,11 @@ add_to_user_searchpath <- function(name, FN) {
     assign(name, FN, 'jupyter:irkernel')
 }
 
+add_to_base_searchpath <- function(name, FN) {
+    unlockBinding(name, baseenv())
+    assign(name, FN, baseenv())
+}
+
 get_shadowenv <- function() {
     as.environment('jupyter:irkernel')
 }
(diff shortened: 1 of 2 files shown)


flying-sheep:
i’d say that `add_to_base_searchpath` is a wrong name. it’s `replace_in_base_namespace`.

what’s the difference between `baseenv()` and `.BaseNamespaceEnv`?

jgbradley1:
Renamed the function as per your suggestion. As for the different between `.BaseNamespaceEnv` and `baseenv()`, paraphrasing from [this](http://adv-r.had.co.nz/Environments.html) page

> ...every package has two environments associated with it: the package environment and the namespace environment. The package environment contains every publicly accessible function, and is placed on the search path. The namespace environment contains all functions (including internal functions), and its parent environment is a special imports environment that contains bindings to all the functions that the package needs. Every exported function in a package is bound into the package environment, but enclosed by the namespace environment.


And from the docs,
> .BaseNamespaceEnv is the namespace environment for the base package. The environment of the base package itself is available as baseenv()

Since `readline` and `quit` are public functions, we probably don't care about the internal functions of the base package. Overriding the function assignment will work using either `.BaseNamespaceEnv` or `baseenv()`. It's probably safer to use `baseenv()` though.

flying-sheep:
ah, great, thank you for your research! we already [use `.BaseNamespaceEnv`](https://github.com/IRkernel/IRkernel/search?utf8=%E2%9C%93&q=.BaseNamespaceEnv&type=), and i think correctly.

`baseenv()` seems like the more fitting alternative for our use case!

jgbradley1:
Thanks for the help! I made the changes and all tests have passed now.

flying-sheep:
hmm, it’s probably hard to add a test for interactive stuff, so i’m just gonna merge it!

flying-sheep:
thank you for your contribution!

PR #475, opened 2017-05-16
add user-specified rprofile

jburos:
Minor pull request to add an R_PROFILE_USER option to the IRkernel spec. 

In our use case, we often have project-level `.Rprofile` files & want to maintain separate kernels for each project.

### DIFF
--- a/R/installspec.r
+++ b/R/installspec.r
@@ -7,6 +7,7 @@
 #' @param user         Install into user directory (\href{https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html}{\code{$XDG_DATA_HOME}}\code{/jupyter/kernels}) or globally?
 #' @param name         The name of the kernel (default "ir")
 #' @param displayname  The name which is displayed in the notebook (default: "R")
+#' @param rprofile_file Path to kernel-specific Rprofile (defaults to system-level settings)
 #' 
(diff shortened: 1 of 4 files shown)


flying-sheep:
looks good, thank you!

could you remove the `_file` from the parameter’s name? it’s unambiguous anyway and we don’t have to reformat the roxygen `@param` block to keep it aligned properly.

flying-sheep:
huh, the tests fail on mac…

why though? any idea? or do you think that’s a travis change that caused it?

jburos:
I noticed that too .. I didn't change anything there but will investigate

flying-sheep:
i just restarted a mac job that previously passed: https://travis-ci.org/IRkernel/IRkernel/jobs/218015784

if it fails too, it’s proven that this can’t be it. then i’ll just pull this and investigate the mac failures later.

flying-sheep:
OK, thank you very much!

Restarting the build made it fail the same way, so it definitely is something that changed on Travis’ side.

jgbradley1:
Did you figure out why the tests are failing? I'm running into issues trying to get them to pass on a [new](https://github.com/IRkernel/IRkernel/pull/478) pull request.

flying-sheep:
nope. i’ll just pull everything that has only OSX tests failing until i know

Issue #460, opened 2017-04-05
getPass is not shaded correctly


jgbradley1:
Not sure how this should best be handled but shading getPass by adding it to the `jupyter:irkernel` environment does not properly shade the function when the getPass package is attached.

**Test that fails**
```r
library(getPass)
x <- getPass('Enter password:')
```
getPass() only works when the getPass package has not been attached.

**Possible Solution**
If getPass is assigned to the global environment, then it works as expected when explicitly attached via `library()`.
```r
assign('getPass', .self$get_pass, globalenv())
```

flying-sheep:
we could try to load the `getPass` namespace and replace it there if it’s available, else attach it as we do now.

the global env isn’t ideal because people like to do `rm(list = ls())` or similar.

PR #478, opened 2017-05-28
Fixed issue #460

jgbradley1:
This fix addresses issue #460. I believe it shades the getPass package correctly and also provides an easy way to overwrite functions for other packages in the future.

As a consequence of this change, the following code will error out whereas it worked before.

```r
pass <- getPass("Enter password")
```

Since the getPass function is actually derived from a package and is not made up, notebook developers should have to explicitly attach the package. The following code will work with this fix
```r
library(getPass)
pass <- getPass("Enter password")
```

This change has the added benefit of making R notebooks more portable when users want to download a notebook as a .R script instead of the .ipynb format.

### DIFF
--- a/R/environment_shadow.r
+++ b/R/environment_shadow.r
@@ -7,7 +7,7 @@ add_to_user_searchpath <- function(name, FN, pkg = NULL) {
         assign(name, FN, 'jupyter:irkernel')
     } else {
         env <- getNamespace(pkg)
-        unlockBinding(name, env)
+        .BaseNamespaceEnv$unlockBinding(name, env)
         assign(name, FN, envir = env, inherits = TRUE)
         lockBinding(name, env)
     }
--- a/R/environment_shadow.r
+++ b/R/environment_shadow.r
@@ -2,8 +2,15 @@
 # This is needed to build in our own needs, like properly shutting down the kernel
 # when `quit()` is called.
 
-add_to_user_searchpath <- function(name, FN) {
-    assign(name, FN, 'jupyter:irkernel')
+add_to_user_searchpath <- function(name, FN, pkg = NULL) {
+    if (is.null(pkg)) {
+        assign(name, FN, 'jupyter:irkernel')
+    } else {
+        env <- getNamespace(pkg)
+        unlockBinding(name, env)
+        assign(name, FN, envir = env, inherits = TRUE)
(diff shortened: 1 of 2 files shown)


jgbradley1:
I'm fairly new to travis test logs. It does not look like my code is producing an error. The tests are timing out due to long compilation time (it's hard to tell if an error exist or not
). Is this a [problem](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) with the travis configuration?

flying-sheep:
hi! thanks for your contribution

the last line in the log before the timeout message is

> Running ‘testthat.R’

so it actually hangs while running tests, not while compiling anything.

can you locally run the tests fine? (<kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>T</kbd> in RStudio)

R/environment_shadow.r
```diff
@@ -2,8 +2,15 @@
 # This is needed to build in our own needs, like properly shutting down the kernel
 # when `quit()` is called.
 
-add_to_user_searchpath <- function(name, FN) {
-    assign(name, FN, 'jupyter:irkernel')
+add_to_user_searchpath <- function(name, FN, pkg = NULL) {
+    if (is.null(pkg)) {
+        assign(name, FN, 'jupyter:irkernel')
+    } else {
+        env <- getNamespace(pkg)
```
flying-sheep: what happens if the package isn’t installed?

please make sure this case is handled by falling back to the `'jupyter:irkernel'` environment.

It would also be great if you could add tests for `add_to_user_searchpath` that cover all three usages (without `pkg`, with existing `pkg`, and with non-existing `pkg`

flying-sheep:
OK cool! i’m not really a fan of doing control flow by exception handling.

i.e. instead of relying on `getNamespace(NULL)` throwing an error, i’d prefer if you did:

```r
available <- requireNamespace(pkg, quietly = TRUE)
if (available) {...} else {…}
```

jgbradley1:
I agree with the more defensive approach. I had come up with something similar but your approach tests for the side cases in a simpler manner.

```r
if(any(pkg %in% .packages(all.available = TRUE))) {...} else {...}
```

flying-sheep:
great! travis still only fails for macs, so :shipit:

Issue #428, opened 2016-12-01
Feature request: option to specify full path with installspec


timothymillar:
We're using the IRkernel with Jupyterhub installed inside an anaconda virtual environment.

It would be convenient to install the kernelspec directly into the environment dir (this makes it available globally) e.g:

```
Rscript -e "IRkernel::installspec(prefix = '/path/to/anacondaEnvironment/share/jupyter/kernels/ir3.3.0', displayname = 'R-3.3.0')"
```

This can sort of be achieved by specifying the full path with the `name` argument but is not the intended behavior  (see #427).

takluyver:
Assuming we expose the options from Jupyter directly, the call would look like:

```r
installspec(prefix='/path/to/anacondaEnvironment', name='ir3.3.0', displayname='R 3.3.0')
```

The `share/jupyter/kernels` part of the path is added in automatically.

flying-sheep:
so what we do now (and what will be the `prefix` default) is “determine via calling `jupyter`”, right?

takluyver:
Yeah, `jupyter kernelspec install` has a `--prefix` option to specify where to install to. I'd just pass through to that. If it's not specified, it picks a default location.

`prefix` and `user` are mutually exclusive (or setting both means one is ignored).

timothymillar:
added in  #512

PR #512, opened 2017-10-16
add prefix option to installspec

timothymillar:
Addresses #428 

Allows a kernelspec to be installed into a non-standard location (e.g. an anaconda environment).

By default, the iRkernel specifies `user = TRUE` which conflicts with Jupyters `prefix` option so I have set up `prefix` to ignore the `user` flag if specified. 

Specifying a non-standard location produces the standard Jupyter warning:
```
> IRkernel::installspec(name = 'ir32', displayname = 'R 3.2', prefix='/home/tim/TEST')
[InstallKernelSpec] WARNING | Installing to /home/tim/TEST/share/jupyter/kernels, which is not in ['/home/tim/.local/share/jupyter/kernels', '/home/tim/.miniconda3/share/jupyter/kernels', '/usr/local/share/jupyter/kernels', '/usr/share/jupyter/kernels', '/home/tim/.ipython/kernels']. The kernelspec may not be found.
[InstallKernelSpec] Installed kernelspec ir32 in /home/tim/TEST/share/jupyter/kernels/ir32
> 
```

This should probably include a test but I'm not sure of the best approach so suggestions welcome.

### DIFF
--- a/R/installspec.r
+++ b/R/installspec.r
@@ -8,11 +8,12 @@
 #' @param name         The name of the kernel (default "ir")
 #' @param displayname  The name which is displayed in the notebook (default: "R")
 #' @param rprofile     (optional) Path to kernel-specific Rprofile (defaults to system-level settings)
+#' @param prefix       (optional) Path to alternate directory to install kernelspec into, overrides 'user' (default: NULL)
 #' 
 #' @return Exit code of the \code{jupyter kernelspec install} call.
 #' 
 #' @export
-installspec <- function(user = TRUE, name = 'ir', displayname = 'R', rprofile = NULL) {
+installspec <- function(user = TRUE, name = 'ir', displayname = 'R', rprofile = NULL, prefix = NULL) {
     exit_code <- system2('jupyter', c('kernelspec', '--version'), FALSE, FALSE)
     if (exit_code != 0)
         stop('jupyter-client has to be installed but ', dQuote('jupyter kernelspec --version'), ' exited with code ', exit_code, '.\n')
@@ -32,7 +33,8 @@ installspec <- function(user = TRUE, name = 'ir', displayname = 'R', rprofile = 
     write(toJSON(spec, pretty = TRUE, auto_unbox = TRUE), file = spec_path)
     
     user_flag <- if (user) '--user' else character(0)
-    args <- c('kernelspec', 'install', '--replace', '--name', name, user_flag, file.path(tmp_name, 'kernelspec'))
+    location <- if (is.null(prefix)) user_flag else c('--prefix', prefix)
+    args <- c('kernelspec', 'install', '--replace', '--name', name, location, file.path(tmp_name, 'kernelspec'))
     exit_code <- system2('jupyter', args)
     
     unlink(tmp_name, recursive = TRUE)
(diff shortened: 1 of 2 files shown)


flying-sheep:
Hi, thanks for your contribution.

I’d prefer if conflicting options would result in an error instead of one having higher priority.

maybe you could change the default to `user = NULL`, mention the default in the parameter docs, and `stop('...')` if not at least one of `user` and `prefix` is `NULL`.

timothymillar:
Thanks for the quick response.

I agree that the conflicting options should error, but didn't want to change the current behavior.

Another option would be to leave `user=TRUE` but not prioritise `prefix` so that you could still do a user install with: 
```IRkernel::installspec(name = 'ir32', displayname = 'R 3.2')``` 
but have to do a prefix install with: 
```IRkernel::installspec(user=FALSE, prefix='/path/to/env', name = 'ir32', displayname = 'R 3.2')``` 
which is probably not used that often. 

 
Also, If I set `user=FALSE` and `prefix=FALSE`, Jupyter tries to do a global installation which I guess can be used with `sudo Rscrip ...` so we probably don't want to enforce `user` or `prefix`: 
```
> IRkernel::installspec(name = 'ir32', displayname = 'R 3.2')
[Errno 13] Permission denied: '/usr/local/share/jupyter'
Perhaps you want to install with `sudo` or `--user`?

```

**Comment by flying-sheep** [comment]
> I agree that the conflicting options should error, but didn't want to change the current behavior.

neither does my idea. instead of:

```r
installspec <- function(user = TRUE, ...) {
    exit_code <- system2(...)
    ...
}
```

you just do:

```r
installspec <- function(user = NULL, ...) {
    if (is.null(user)) user <- TRUE
    exit_code <- system2(...)
    ...
}
```

timothymillar:
Just to clarify, if both values default to `NULL` we have to handle the following situations:

- `user=NULL, prefix=NULL` which is treated as `user=TRUE, prefix=NULL` to preserve current behavior 
- `user=TRUE, prefix=NULL` a "user" install
- `user=NULL, prefix='/path/...'` a "prefix" install
- `user=FALSE, prefix='/path/...'` also a "prefix" install
- `user=TRUE, prefix='/path/...'` invalid so caught by `stop(...)` 
- `user=FALSE, prefix=NULL` a "global" install

which can be handled with:

```
if (is.null(user) & is.null(prefix)) user <- TRUE
if (is.null(user) & !is.null(prefix)) user <- FALSE
if (user & !is.null(prefix))
    stop('"user" and "prefix" are mutually exclusive')
...

user_flag <- if (user) '--user' else character(0)
prefix_flag <- if (!is.null(prefix)) c('--prefix', prefix) else character(0) 
args <- c('kernelspec', 'install', '--replace', '--name', name, user_flag, prefix_flag, file.path(tmp_name, 'kernelspec'))
```

flying-sheep:
exactly. although

```r
if (is.null(user) & is.null(prefix)) user <- TRUE
if (is.null(user) & !is.null(prefix)) user <- FALSE
```

can be condensed to

```r
if (is.null(user)) user <- is.null(prefix)
```

flying-sheep:
great!

can you now please regenerate the documentation [to make travis happy](https://travis-ci.org/IRkernel/IRkernel/jobs/291418688#L1352)?

timothymillar:
Should be fixed now, ran into [this](https://github.com/klutometis/roxygen/issues/595)

flying-sheep:
great, thank you!

PR #523, opened 2017-11-20
Message protocol 5 doesn't use 'source' field

takluyver:
http://jupyter-client.readthedocs.io/en/latest/messaging.html#display-data

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,6 @@ matrix:
     - r: release
       env: DEPS=github
       os: linux
-    - r: release
-      env: DEPS=cran
-      os: osx
-    - r: release
-      env: DEPS=github
-      os: osx
     - r: devel
       env: DEPS=cran
       os: linux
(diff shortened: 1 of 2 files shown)


flying-sheep:
Yeah, they’ve been failing for some time, but I have no idea why.

Could you please add a test for protocol validity? (or enable checking this in jupyter-kernel-tests, as i assume it should be in there)?

takluyver:
It looks like JKT doesn't currently support checking for unexpected extra attributes. I'll work on that, but it's not a five minute job to do it right.

Are you happy if I comment out the Mac tests and open an issue for someone to look at them later? They're slower than the Linux tests, and CI is less useful if we're used to it failing.

flying-sheep:
yes, very happy, i wanted to do that long ago. no idea why i didn’t…

kojix2:
Thank you @takluyver @flying-sheep

PR #564, opened 2018-05-23
Similar to https://github.com/IRkernel/IRkernel/pull/523 execution  count is causing problem when saving R notebook with an output cell containing error message.

ckbhatt:


ckbhatt:
@flying-sheep @takluyver please review

flying-sheep:
i had problems when leaving out the execution count somewhere…

@takluyver you know the protocol better than me: can you tell us when to include and when to omit it?

takluyver:
Off the messages sent through the iopub channel, I think it is only `execute_input` (which echoes the submitted code for other frontends) and `execute_result` (used in Python when you have a bare expression at the end of a cell) that should have an execution count attached.

On the shell channel, the `execute_reply` message also has it.

Docs here: http://jupyter-client.readthedocs.io/en/latest/messaging.html#messages-on-the-iopub-pub-sub-channel

ckbhatt:
https://github.com/jupyter/nbformat/blob/master/nbformat/v4/nbformat.v4.schema.json#L327
does not allow additional properties and does not include execution_count
Here is a message that I am getting while validating notebook content
`[E 2018-05-23 22:58:37.801 AzureFS azuremanager:228] Error while saving model [/Untitled7.ipynb]: HTTP 400: Bad Request (Error saving notebook: Notebook validation failed: Additional properties are not allowed ('execution_count' was unexpected):
    {
     "ename": "ERROR",
     "evalue": "Error in eval(expr, envir, enclos): object 'hsjh' not found\n",
     "traceback": [
      "Error in eval(expr, envir, enclos): object 'hsjh' not found\nTraceback:\n"
     ],
     "execution_count": 1,
     "output_type": "error"
    })`
Does that mean though IRKernal is complying with spec, nbformat is not?

takluyver:
Yeah, I'm pretty sure you're right that the `error` message type shouldn't have an `execution_count` field.

@flying-sheep feel free to merge if the implementation looks OK to you.

flying-sheep:
OK, thanks all!

kojix2:
Thank you IRkernel developer!!

Issue #567, opened 2018-06-10
Catching q() in IRkernel


PeterRuckdeschel:
I was reported the following error thrown from R (used in IRkernel) when using my contributed R package distr

```
distr::Beta(3, 2) + distr::Beta(3, 2)

Error in switch(save, default = , yes = TRUE, no = FALSE, ask = ask("Save workspace image? [y/n/c]: "), : EXPR must be a vector of length 1

Traceback:

1. distr::Beta(3, 2) + distr::Beta(3, 2)
2. distr::Beta(3, 2) + distr::Beta(3, 2)
3. getLow(e1)
4. getLow(e1)
5. .local(object, ...)
6. q(object)
7. q(object)

``` 

My guess is that this comes from IRkernel catching calls to q() and evaluating them in a non-R-standard manour.

Background: 
In our R-package distr, we make q() an S4 generic, [i.e., a function which may digest varying input and react on it according to the class of the first argument(s), like with print or plot (albeit those are S3 generics...)].   Making q() an S4 generic is completely in line with R semantics. Now, after loading the namespace of distr, if the first argument of q (which is controlling method dispatch) is an S4 object of class "distribution" q() returns the quantile function of this distribution. This is what q(object) does in getLow() in the code snippet above. Otherwise, i.e., if the first argument of q() is not a distribution object, it falls back to standard R, calling function base::q() (i.e., q() from R's base package) as usual, which is the function used for quitting R, so you get the seen error message... 

For new code involving calls to q(), we already have an alias q.l() for q() in our package distr for quite a while, which would fix this issue in new code.  But it seems that the IRkernel non-standard evaluation of q() even goes through code protected in namespaces. This seems to throw the error in the code snippet, as q(object) is contained in the internal definition of getLow(), and altering lots of existing code is not really an option for us.  

So I would ask you to also mimic the S3/S4 semantics for q(), i.e., check whether q() has been made an S3/S4 generic, and if so only to use the non-standard evaluation if it dispatches to using base::q() -- because otherwise it will break code at quite some unforeseeable places.

Thanks already, and best regards, Peter Ruckdeschel

flying-sheep:
I see, sounds reasonable. Wanna do a PR?

PeterRuckdeschel:
just some minimal testing example, if you are digging into this:

setClass("foo", representation=representation(x="numeric"))
x <- new("foo")
setGeneric("q", function(save = "default", status = 0, runLast = TRUE) standardGeneric("q"))
[...]

PR #570, opened 2018-06-14
Internally use original `quit` function instead of the one provided for users

### LINKED ISSUE #569: Message-based shutdowns are ignored after at least one cell has executed
Hello, 
We're experiencing an issue where the IRkernel is not gracefully shutdown once at least one cell has executed.  In this case, the replaced `quit()` function is called which doesn't stop the kernel, so the loop processing continues.  This triggers the jupyter framework to enter its kill sequence since the kernel is still considered running - ultimately leading to a harsh death.

In environments, like [Jupyter Enterprise Gateway](https://github.com/jupyter-incubator/enterprise_gateway), where kernels are running as managed resources on remote nodes, this kill sequence leads to application managers terminating the kernel and marking the application as KILLED (e.g., in YARN and IBM Spectrum Conductor) when, in reality, they should be marked as FINISHED - since its clearly the kernel user's choice to shut down.  In addition, having to incur this extra override is expensive and not always reliable.

To reproduce this, you can enable jupyter debug and kernel tracing:
```
docker run -it --rm \
    -e JUPYTER_LOG_LEVEL=3 \
	-p 8888:8888 \
	-v `pwd`:/src_irkernel \
	${DEV_IMAGE} bash -c 'R CMD INSTALL -l /opt/conda/lib/R/library /src_irkernel && \
	                        jupyter notebook --no-browser --port 8888 --ip='*' --log-level=10'
```

Here's the pertinent output.  In this case a cell has executed (e.g., `1+1`).  You can see the main loop is re-entered and the jupyter framework has determined that it must invoke its kill sequence:
```
DEBUG: main loop: after poll. ZMQ code: 1; Errno: 11
DEBUG: main loop: control
DEBUG: Control: beginning
DEBUG: Control: recv msg of type shutdown_request
DEBUG: Control: shutdown...
DEBUG: Sending msg shutdown_reply
DEBUG: main loop: beginning
[D 15:59:28.652 NotebookApp] Kernel is taking too long to finish, killing
[I 15:59:28.659 NotebookApp] Kernel shutdown: 46f03728-d6bd-41b5-a74f-d7d160b6ca54
```
vs. when no execution has occurred, the main loop is not re-entered and the kernel has been gracefully shutdown (no kill sequence invoked):
```
DEBUG: main loop: after poll. ZMQ code: 1; Errno: 11
DEBUG: main loop: control
DEBUG: Control: beginning
DEBUG: Control: recv msg of type shutdown_request
DEBUG: Control: shutdown...
DEBUG: Sending msg shutdown_reply
[I 15:59:04.015 NotebookApp] Kernel shutdown: abeb6635-ab73-4ec1-b6db-0a26a0e7e92b
```

This appears related to #153 and #196 (and #567 as well perhaps).  Unfortunately, I don't know R well enough to help with any substantial change.  If I comment out the code that replaces the base package quit function, all is good - although I'm not sure what is now broken.  At a minimum, some kind of environment variable support to indicate that replacement not be performed would be helpful.

Thank you.

kevin-bates:
The replacement of the base `quit()` function with `execution.r/quit()`
prevents message-based shutdown requests from completing, triggering the
juptyer framework's kill sequence.  If the shading is not performed,
kernels can be properly shutdown via the initial request and prevent
the additional overhead and potential unreliability of kill sequence
(signals, REST apis, etc.).

Since my knowledge of R is extremely limited, there may be far better
ways to handle this, so please feel free to comment to that affect.
Also, it's not clear to me what functionality I'm preventing by making
this change.

To enable the skipping of the replacement, a new option has been added:
`jupyter.replace_quit` (env: `JUPYTER_REPLACE_QUIT`) default value is
`TRUE` so the current functionality is retained.  When set to FALSE,
the replacement of the base `quit()` method is not performed.

I'm not thrilled by the name either, and if I better understood why the
`quit()` method was necessary in the first place, I'd be happy to come
up with a different name.  I kinda wonder if using `jupyter.in_kernel`
for this would be better since it seems like we'd always want successful
message-based shutdown requests.

Please refer to Issue #569 for how to test.  The key thing is that,
at a minimum, Notebook DEBUG logging be enabled so one can determine
if the kill sequence is being invoked.  Another way is if the kernel's
termination takes on the order of 5 seconds vs sub-second.

Fixes #569

### DIFF
--- a/R/environment_shadow.r
+++ b/R/environment_shadow.r
@@ -36,6 +36,7 @@ backup_env$utils_flush_console <- function(...) {}
 init_backup_env <- function() {
     backup_env$base_flush_connection <- base::flush.connection
     backup_env$utils_flush_console <- utils::flush.console
+    backup_env$base_quit <- base::quit
 }
 
 # Adds functions which do not need any access to the executer into the users searchpath
--- a/R/kernel.r
+++ b/R/kernel.r
@@ -363,7 +363,8 @@ handle_control = function() {
 shutdown = function(request) {
     send_response('shutdown_reply', request, 'control', list(
         restart = request$content$restart))
-    quit('no')
+    # Always call the base quit() during shutdown since execution shadows it.
+    backup_env$base_quit('no')  # bound during startup in .onLoad
 },
 
 initialize = function(connection_file) {
@@ -407,6 +408,7 @@ initialize = function(connection_file) {
 
 run = function() {
     options(jupyter.in_kernel = TRUE)
+
     while (TRUE) {
         log_debug('main loop: beginning')
         r <- tryCatch(


takluyver:
Thanks! I hope we can find a way to fix the underlying problem, though, rather than just letting users choose between two different problems.

takluyver:
I think that the replacement needs to be done in some form, because if a kernel process just exits by itself, Jupyter will think it has died and automatically restart it. To exit in Jupyter, the code needs to send a message to the frontend asking to be shut down. I know that's somewhat convoluted, but that's the way it goes.

There's another issue open with a separate problem with the override - #567.

kevin-bates:
OK - thanks.  I thought the kernel's shutdown was triggered by the notebook user (via File/Close and Halt) or the REST API issuing a DELETE request  - which then sends a shutdown_request via the control port?  

Are you trying to support the case where the user issues quit() within a cell to terminate the kernel?

takluyver:
You can also use the R kernel from other Jupyter frontends, like the Qt console. In a console, it's not unusual to issue an exit/quit command from inside the kernel to quit both the frontend and the kernel together.

takluyver:
I think the answer will be something like: when we replace the `quit` function, keep the original quit function around somewhere, and use that when handling `shutdown_request`. I'm not familiar enough with R to know the best way to do that, though.

kevin-bates:
@takluyver - I've gone ahead and amended the original commit since none of those changes apply and found a way to bind the `base::quit` function to a new name (original_quit) that can then be called during shutdown, avoiding the call to the custom (shadowed) quit function.

flying-sheep:
hey, thanks for the contribution! keeping the original version of functions around already happens in IRkernel [here](https://github.com/IRkernel/IRkernel/blob/1cd737d9fff91ca3b3415698b353b569bbb8df3e/R/environment_shadow.r#L36-L39). i suggest you use the same approach for `quit`!

kevin-bates:
@flying-sheep - thank you for the suggestion.  Unfortunately, I don't know R well enough to take that same approach.  I've made a couple attempts, and all fail either at build or run time - seems like scoping is my issue with those.  I will need some help from someone more knowledgable in R in order to take the `new.env()` approach.

flying-sheep:
well, you could simply put a new line into `init_backup_env` mirroring the lines that are already there, like:

```r
backup_env$base_quit <- base::quit
```

and then call `backup_env$base_quit('no')` in `shutdown`.

kevin-bates:
@flying-sheep - thank you.  I had figured by `approach` you had meant to do things in a similar way - but interpreted that at too high a level.  I suppose `onLoad` is like an initialization method but I'm still not sure how things know about `backup_env`.  At any rate, your suggestion works great and I've amend the commit.

flying-sheep:
Great!

the `backup_env` is just a package-level variable (i.e. it lives in the same namespace as package-internal/unexported functions). so every function in the IRkernel package can access that environment.

as you correctly guessed, `.onLoad` is the package initialization method. it gets executed by R automatically when a package defining it gets loaded.

kevin-bates:
Awesome. Thank you for your help @flying-sheep and @takluyver.

When is the next IRkernel release planned?

flying-sheep:
No plans, but we can always tag a new one. Is there some important change that’s missing?

kevin-bates:
@flying-sheep - yes, we have a customer that raised this issue in the first place and we will be updating our [docs](https://jupyter-enterprise-gateway.readthedocs.io/en/latest/getting-started-kernels.html#installing-support-for-r-irkernel) to include the tag (at least until the next release is produced).  Thank you.

Fincer:
I would like to test this solution to solve similar (or same) issue described [here](https://github.com/IRkernel/IRkernel/issues/582).

I have deployed and configured JupyterHub service and IRkernel is successfully installed and works well, too. However, I am not too familiar with R language and therefore, the instructions in the previous posts are somewhat vague to me. Could you provide bit more specific instructions how to implement the described solution to an existing JupyterHub/IRKernel installation? Is there, for example, any arguments or parameters I could use in `jupyter/kernels/ir/kernel.json` file or should I look to somewhere else to test the solution?

Not sure if relevant but user environments ([JupyterLab](https://github.com/jupyterlab/jupyterlab) in my case) are initialized with [sudospawner](https://github.com/jupyterhub/sudospawner) at the moment.

kevin-bates:
Hi @Fincer - the changes described above are, to the best of my knowledge, specific and isolated to the IRKernel code.  The issue was that one form of the `quit()` function was shadowed by another and, when completely shutting down the kernel, we needed to use the original function.   There is no external exposure of this functionality (i.e., its all completely within the IRKernel package and cannot be "controlled" via  user-facing options, like those that would be exposed to `kernel.json`).

If you can describe your particular scenario, that might help others help you.  This particular issue was 5 years ago, and I've hardly touched R code since then, and, as noted in the original description, had hardly touched R code prior.  As a result, I'm not sure how much assistance I can be anyway.

Fincer:
Hey @kevin-bates , and thanks for the quick reply!

Well, if the changes are completely internal and not configurable via any settings (e.g. in `kernel.json`), it seems there is not much to do then. I just thought the quit functionality could be configured somewhere and you could know the necessary details for setting things up properly.

I may open a new issue to explain my case in detail. Thanks again!

flying-sheep:
Hmm, strange, the kernel calls `base::quit('no')` on shutdown, so there should be no need to call it:

https://github.com/IRkernel/IRkernel/blob/1eddb304b246c14b62949abd946e8d4ca5080d25/R/kernel.r#L346

Fincer:
Yeah, this might be a configuration issue or a bug related not to IRKernel but to JupyterLab/sudospawner, instead. I still need to do more investigation. It's still good to narrow down potential culprits, though.

PR #606, opened 2019-01-08
Handle repr.plot.res; fixes #458

### LINKED ISSUE #458: repr.plot.res does not work as expected
I found repr.plot defaults [here](https://github.com/IRkernel/repr/blob/master/R/options.r#L51-L61), including res, which looks to be dots per inch (dpi). However, when I raise the res from 120 to 300, the plot gets bigger, despite me setting height and width, instead of staying the same size and getting crisper, as I would hope.

========================================

More detail from [my stack overflow question](http://stackoverflow.com/questions/43123034/how-to-get-high-resolution-plots-in-a-jupyter-notebook-using-the-r-kernel):

Using SVG in a Jupyter notebook with R kernel mangles the text. See [here](http://compbio.ucsd.edu/outputting-beautiful-jupyter-notebooks-r-kernel-edition/#html-output-mangled-graphics-text) and [here](https://github.com/jupyter/nbconvert/issues/129). The workaround is to use PNG. It's a bit blurry. In Python, you can ask the PNG renderer to use more dots per inch with

```InlineBackend.figure_format = 'retina'```

See for example [trick 15 here](https://www.dataquest.io/blog/jupyter-notebook-tips-tricks-shortcuts/).

How can I get higher resolution plots using the R kernel? SVG workaround? PNG renderer settings?

I'm using ggplot in R.

### PR DESCRIPTION
**Pull request by raivivek**

### CODE REVIEW

**Comment by flying-sheep** [comment]
Ah, nice idea! Thank you!

Those attributes are used here:

https://github.com/IRkernel/IRkernel/blob/45dad16502694679900d2af4bfc14b82d646ae27/R/execution.r#L171

Do you think you could (instead of premultiplying) add another attribute (`.irkernel_res`) and use it in that line?

```r
formats[[mime]] <- mime2repr[[mime]](
    plotobj,
    attr(plotobj, '.irkernel_width'),
    attr(plotobj, '.irkernel_height'),
    res = attr(plotobj, '.irkernel_res')
)
```

`mime2repr[[mime]]` will be one of the functions in [here](https://github.com/IRkernel/repr/blob/master/R/repr_recordedplot.r), and each of them that doesn’t have a `res` parameter will accept and ignore it when passed (SVGs don’t have a resolution!)

raivivek:
I was wondering if you could help me understand a few things. Right now, there's no `.irkernel_res` attribute being set, so does that mean setting `option(repr.plot.res = XX)` right now has no effect? I guess the question is; if `res` is being passed to underlying devices and they are not respect it or IRkernel is not doing it right?

flying-sheep:
There are few silly questions, and this is not one of them!

Setting those attributes helps in one case: If there’s another `options` call being made after the plot call in a cell:

considering every code block is a cell: this works perfectly

```r
options(repr.plot.res = 80)
plot(...)
```

```r
options(repr.plot.res = repr_option_defaults$repr.plot.res)
```

This however doesn’t:

```r
options(repr.plot.res = 80)
plot(...)
options(repr.plot.res = repr_option_defaults$repr.plot.res)
```

Neither does this:

```r
for (res in 40*(2:4)) {
    options(repr.plot.res = res)
    plot(...)
}
```

After adding the attribute and using it in `send_plot`, all of the above will work.

R/execution.r
```diff
@@ -223,8 +229,15 @@ handle_graphics = function(plotobj) {
         send_plot(last_recorded_plot)
     }
     # need to be set here to capture the size and have it available when the plot is sent
-    attr(plotobj, '.irkernel_width')  <- getOption('repr.plot.width',  repr_option_defaults$repr.plot.width)
-    attr(plotobj, '.irkernel_height') <- getOption('repr.plot.height', repr_option_defaults$repr.plot.height)
```
flying-sheep: Why are you changing these lines at all?

You just need to add a `attr(plotobj, '.irkernel_height') <- ...` line, and the change above (`mime2repr[[mime]](..., res = attr(plotobj, '.irkernel_res'))`) will make things look correct, right?

flying-sheep:
There’s two parts: make things work in a loop. my comments were about this, and this is now handled.

The second part is the semantics of res itself.

- `res` describes PPI (pixels per inch). PPI is a relation of physical size and pixel size. Plotting to any raster graphics device (like `png`) creates an image with `width_in_pixels = width_in_inch * PPI`. `.irkernel_width` is the width in inch and `.irkernel_res` is PPI, so *the current code is correct*.

- Screens also have a PPI value, and high-res sceens have two of them:

    1. Real PPI, the amount of pixels that are in an inch of monitor horizontally or vertically
    2. CSS PPI, what’s meant by `1px` in CSS and HTML. *This is often half of the real PPI*, so `width: 100px` means “set width to 100 CSS pixels / 200 screen pixels”.

    If you display an `` in HTML without specifying dimensions, it will use its intrinsic size (e.g. 400px×300px) as size in CSS pixels and look blurry. Therefore, to make it look crisp on high-res screens you need to specify e.g. half of the CSS pixels for width and height, in our example ``.

Therefore you need to

1. save the size in inches and the PPI into the attributes unchanged so that plots with the right amounts of pixels are created in `send_plot` (*I edited this PR so that it does that*)
2. to get the real_width/display_width factor, either a) add an option or b) derive it from the PPI value. (optimally, add the option and allow to set it to `NA` which means “derive”)
3. if the factor is not 1, set the real width as [metdatda](https://nbformat.readthedocs.io/en/latest/format_description.html#display-data). the metadata is already modified in `send_plot`, so it should be easy to figure out what to do.

flying-sheep:
I think I handled this in 44a92feba841e9959788f3c6750c73d293ca9387

PR #651, opened 2020-02-05
Fix string completion test

whatalnk:
It seems that "testthat.Rout" file is created when testing.


flying-sheep:
Thank you! Locally it doesn’t work that way, but I’ll figure this out as soon as I run into it next time :smile:

PR #655, opened 2020-03-30
Add sys_prefix option to mirror --sys-prefix

rkdarst:
In my work deploying jupyter to people, I've been annoyed that there
is no `--sys-prefix` equivalent for IRkernel.  Well, maybe I should do
something about it...

- Add installspec(sys_prefix=TRUE) option to install using the
  equivalent of --sys-prefix.  Support added in jupyter-client 4.3
  (2016).  This is passed directly to the underlying `jupyter
  kernelspec install` command.
- No tests yet, but user=TRUE doesn't seem to be either...

I'm not an R expert, and I haven't even tested this code myself.  But
I figure that making prototype quickly is better than abstract
discussion, so here it is.  This would have saved me plenty of time
over the years.

Whether or not this goes through, I will try to improve the
installation docs some - it would have also saved me time, as a
cluster manager who needs to know some more details about what's going
on.

### DIFF
--- a/R/installspec.r
+++ b/R/installspec.r
@@ -9,19 +9,20 @@
 #' @param displayname  The name which is displayed in the notebook (default: "R")
 #' @param rprofile     (optional) Path to kernel-specific Rprofile (defaults to system-level settings)
 #' @param prefix       (optional) Path to alternate directory to install kernelspec into (default: NULL)
+#' @param sys_prefix   (optional) Install kernelspec using the "--sys-prefix" option of the currently detected jupyter (default: NULL)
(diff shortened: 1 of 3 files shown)


rkdarst:
OK, I managed to fix the doc issue, and added some more docs about it.  In the process I found a syntax error in the R file, which is now fixed.

So... any other suggestions?  I still haven't given this any tests at all... I'm not that much of an R user!  Maybe I can ask someone to help with that...

flying-sheep:
Looks good, thank you!

PR #683, opened 2021-05-03
adjust comps regex not to wrap names with leading . in backticks

### LINKED ISSUE #681: Autocomplete wraps valid identifiers in backticks
I am trying to enter `.Last.value` and IRKernel always wraps it in backticks, which is unnecessary.

I ran this to see what `comps` is just before the backtick line is run:

https://github.com/IRkernel/IRkernel/blob/a862774dc0758b23ef4e911a3c0b1d9e550652bc/R/completion.r#L52

```
trace(IRkernel:::fixup_comps, quote(dput(comps)), at = 7)
```

Then type `.L`<kbd>Tab</kbd> and see in the logs that `comps` is:

```
comps <- c(".Last.updated", ".Last.value", ".Library", ".Library.site")
```

And indeed, these comps get wrapped:

```
gsub("^([_.].*?|.*?[^\\w\\d._].*?|.*?[.]\\d)$", "`\\1`", comps, perl = TRUE)
# [1] "`.Last.updated`" "`.Last.value`"   "`.Library`"      "`.Library.site`"
```

Unfortunately I don't really understand what the regex is trying to do... FWIW I have written a regex for identifiers in R before:

```
.IsValidRName <- function(x) {
   grepl("^([a-zA-Z][a-zA-Z0-9._]*|[.]|[.][a-zA-Z._][a-zA-Z0-9._]*)$", x)
}
```

MichaelChirico:
Closes #681

This passes the existing tests mentioned in the issue, however PTAL as I'm not sure how comprehensive that list is.

### DIFF
--- a/R/completion.r
+++ b/R/completion.r
@@ -49,7 +49,7 @@ fixup_comps <- function(comps) {
     
     # wrap non-identifiers with ``
     # https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Identifiers
-    comps <- gsub('^([_.].*?|.*?[^\\w\\d._].*?|.*?[.]\\d)$', '`\\1`', comps, perl = TRUE)
+    comps <- gsub('^(_.*?|[.]{2,}.*?|.*?[^\\w\\d._].*?|.*?[.]\\d)$', '`\\1`', comps, perl = TRUE)
     
     # good coding style for completions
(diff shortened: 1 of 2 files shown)


flying-sheep:
great, thank you! sorry for the delay

flying-sheep:
and released!

PR #703, opened 2021-12-08
Short-circuit completion if line is NULL

### LINKED ISSUE #702: Kernel restarting due to an exception in completions (in core R)
- [x] I followed the [official installation instructions](https://irkernel.github.io/installation/)
- [x] I reproduced the error with the newest versions of `IRkernel`, `IRdisplay`, and `repr`
  
- [ ] I included a [minimal reproducible example](https://stackoverflow.com/a/5963610/247482)
- [x] This the right repository: I think the issue is with IRkernel, not
      [IRkernel/repr](https://github.com/IRkernel/repr/issues/new) or a third party repository

So that's hard to reproduce, but I'm getting this a lot in JupyterLab with LSP extension installed. The LSP extension can trigger completions after typing any character, so it may sometimes trigger in odd places and I think this leads to some edge case in completions handling; pretty often the kernel will die and restart with the following message in the console:

```R
Error in suppressWarnings(gregexpr("[^\\.\\w:?$@[\\]]+", linebuffer, perl = TRUE))[[1L]] :
  subscript out of bounds
Calls: <Anonymous> ... handle_shell -> complete -> completions -> <Anonymous>
Execution halted
[I 2021-12-01 15:50:58.092 ServerApp] AsyncIOLoopKernelRestarter: restarting kernel (1/5), keep random ports
kernel 6940fbfd-10d1-4918-ada7-7028d880a2f3 restarted
```

It seems to come from this call:

https://github.com/IRkernel/IRkernel/blob/d872064ff4324e61509af5ebd0038fd16998de89/R/completion.r#L23-L24

To a core function which was not modified in the last 8 years:

https://github.com/wch/r-source/blob/79298c499218846d14500255efd622b5021c10ec/src/library/utils/R/completion.R#L286-L291

I wonder, is this off-by-one error on my side, or potentially something in IRKernel? In either case I thought it would be worth to record an issue here to ask if anyone saw something similar.

I wonder if this could be that sometimes completing on empty `line` triggers this issue?

krassowski:
Fixes #702


flying-sheep:
Sorry for the enormously late reply. Wouldn’t `if (is.null(line)) line <- ''`work better? That way I’d assume you’d get a list of defined variables or so instead of nothing …

flying-sheep:
OK!, thank you for your PR!

li1311139481:
hello, i'm stupid and can you tell me how to use 
the "Re-assign null line to empty string" ?       is run it in jupyterlab'R every time when i want use R in jupyterlab?   Thanks

PR #705, opened 2021-12-10
Get parent header from current request

randy3k:
- Use `current_request` as`parent_msg` is `parent_msg` is `NULL`
- Support `metadata` in `send_response`


randy3k:
It allows something like this.
Screen Shot 2021-12-09 at 11 14 04 PM

buchgr:
Friendly ping @flying-sheep. Would love to see this merged!

PR #734, opened 2023-03-03
improve tab-completion of non-syntactic names

### LINKED ISSUE #733: Autocomplete is too greedy in "backing up" to include too many `$` in a recursive call
```r
xx = list(host = list(list(h1 = 1, h2 = 2), list(h3 = 3, h4 = 4)))
IRkernel:::completions("xx$host[[1]]$h")
# $comps
# [1] "xx$`host[[1]]$h1`" "xx$`host[[1]]$h2`"

# $start
# [1] 0

# $end
# [1] 14
```

Of course, we should only consider the entity after the last `$`. It looks like the completion may be anchoring to the first `$`, but it's strange because it gets the valid completions (`h1`, `h2`) correct.

MichaelChirico:
Closes #733.

Incorporates feedback/suggestions from the upstream discussion for R-core:

 - https://stat.ethz.ch/pipermail/r-devel/2023-March/082388.html
 - https://bugs.r-project.org/show_bug.cgi?id=18479
 - https://github.com/r-devel/r-svn/pull/112

@flying-sheep please guide me how to add test cases. I think, as opposed to the other test cases I see thus far in tests/testthat/test_ir.py, we need to define some R variables first (like `xx` in the #733 issue body or the examples added to the r-svn PR), and I don't see an example of how to do that in the suite as of now.

### DIFF
--- a/R/completion.r
+++ b/R/completion.r
@@ -58,8 +58,9 @@ fixup_comps <- function(comps) {
     last_match <- vapply(lead_matches, tail, n = 1L, integer(1L))
     has_match <- last_match > 0L
     leading <- rep("", length(comps))
-    leading[has_match] <- substr(comps[has_match], 1L, last_match - 1L)
-    comps[has_match] <- substr(comps[has_match], last_match, nchar(comps[has_match]))
+    comps_with_leading <- comps[has_match]
+    leading[has_match] <- substr(comps_with_leading, 1L, last_match - 1L)
+    comps[has_match] <- substr(comps_with_leading, last_match, nchar(comps_with_leading))
     
     # wrap non-identifiers with ``; h/t the related r-devel thread:
     #   https://stat.ethz.ch/pipermail/r-devel/2023-March/082388.html
--- a/R/completion.r
+++ b/R/completion.r
@@ -51,6 +51,9 @@ fixup_comps <- function(comps) {
     comps <- gsub(sprintf('(%s)$', re_trail), '', comps, perl = TRUE)
 
     # split off everything before the last special operator (a la utils:::specialOpLocs)
+    # NB: use look-behind so that we can use the output directly without worrying about
+    #   match.length. Separate look-behind conditions because each one must have a
+    #   fixed length.
     lead_matches <- gregexpr("(?<=[$])|(?<=@)|(?<=[^:]::)|(?<=:::)", comps, perl = TRUE)
     last_match <- vapply(lead_matches, utils::tail, n = 1L, integer(1L))
     has_match <- last_match > 0L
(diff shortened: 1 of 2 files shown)


R/completion.r
```diff
@@ -45,20 +45,32 @@ completions <- function(code, cursor_pos = nchar(code), fixup = TRUE) {
 fixup_comps <- function(comps) {
     # TODO: only do this if we are not in a string or so
     re_trail <- '=|::'
-    re_lead <- '[\\w\\d._]+(?:\\$|@|:::?)'  # TODO: allow foo$`_bar`$baz<tab>
```
MichaelChirico, line 48: R itself doesn't do auto-completion for this case, so not sure there's anything to be done on the IRkernel side:

```r
foo = list(`_bar` = list(bazaar = 1))
foo$`_bar`$baz<tab> # nothing
```

flying-sheep:
> @flying-sheep please guide me how to add test cases. I think, as opposed to the other test cases I see thus far in tests/testthat/test_ir.py, we need to define some R variables first (like `xx` in the #733 issue body or the examples added to the r-svn PR), and I don't see an example of how to do that in the suite as of now.

Sure! Are you solid in Python? If not, I can set things up for you, otherwise please read on.

`jupyter_kernel_test` (the testing library I use) performs the tests in the `completion_samples` array in the [`test_completion`](https://github.com/jupyter/jupyter_kernel_test/blob/69e02cb5fab1c3048f2d0169c4e52cb81cbd71b0/jupyter_kernel_test/__init__.py#L154-L159) function, which only accesses `'matches'` and `'text'`. So there’s no way to directly have setup/teardown code in the `completion_samples` array.

You could duplicate what `test_completion` does:

1. iterate a different array defined e.g. like this:

   ```py
   non_syntactic_completion_samples = [
       dict(text='...', matches='...', vars=dict(xx='"foo"'))
   ]
   ```

2. Duplicate what `test_completions` does, but with setup/teardown:

   ```py
   def test_non_syntactic_completion(self):
       for sample in self.non_syntactic_completion_samples:
	       with self.subTest(text=sample["text"]):
               # +++
               for var_name, var_code in sample["vars"].items():
                   self._execute_code(f'{var_name} <- {var_code}')
               # /+++
	           msg_id = self.kc.complete(sample["text"])
	           reply = self.get_non_kernel_info_reply()
	           validate_message(reply, "complete_reply", msg_id)
	           if "matches" in sample:
	               self.assertEqual(set(reply["content"]["matches"]), set(sample["matches"]))
               # +++
               for var_name in sample["vars"]:
                   self._execute_code(f'rm("{var_name}")')
               # /+++
   ```

Only caveat: The `jupyter_kernel_test` version we use is pretty old, so something might be off. If you run into problems, I’ll update it!

MichaelChirico:
Thanks! I think I've got the basic idea. I'll check the CI result for green.

Please let me know your preference for styling long lines of Python.

MichaelChirico:
Tests passing now 🎉 

Thanks for the help!

flying-sheep:
wonderful, thank you!

Issue #722, opened 2022-11-23
High amount of wrong warning messages when using nondefault font in R Kernel


AlFontal:
## Environment data

-   OS and version: MacOS Monterrey 12.6 on Apple M1 Pro chip.
-   R version: `4.2.2`
-   IRkernel version: `1.3.1`
-   Jupyter server running: Local

## Steps to reproduce

The issue arises whenever using a non-default font with any plotting library (well, at least I have tested both `base` and `ggplot2` plots). In my case I am using the `Georgia` family but any other font shows the same behaviour.

The most minimal example would be the following:

#### `base`

```r
plot(mtcars$wt, mtcars$mpg, family='Georgia')

```
#### `ggplot2`

```r
ggplot(mtcars, aes(x=wt, y=mpg)) + 
    geom_point() +
    theme(text=element_text(family='Georgia'))
```

### Expected behaviour

One would expect the output as seen when running directly in R: No warnings (since the font is actually _found_ and _registered_), and the figures properly rendered:

#### `base`

<img width="594" alt="imagen" src="https://user-images.githubusercontent.com/10958332/203567967-b05d5ef2-9dc4-4f4f-a2bf-16b7e14d50ff.png">

#### `ggplot2`

<img width="594" alt="imagen" src="https://user-images.githubusercontent.com/10958332/203568066-3a0f72d2-01bd-4968-b7b4-9cb635eb1609.png">

### Actual behaviour

When executed in a Jupyter notebook (both in a native Jupyter Notebook or within VS Code which is what I was using), the plots generated are actually correct, and display the same font. However, I get some _really_ long lists of warnings (I suspect one for each single glyph in the figure). 

The kernel is able to load the fonts, so I am not sure of what might cause this. Given how bad the warnings handling is in R, this becomes _really_annoying, as I don't want to simply supress all warnings (nor I want to handle every plot manually...)

I append the full warning logs I get in each case:

<details>
<summary> Show full <b>base</b> output </summary>

```
Warning message in axis(side = side, at = at, labels = labels, ...):
“font family 'Georgia' not found in PostScript font database”
Warning message in axis(side = side, at = at, labels = labels, ...):
“font family 'Georgia' not found in PostScript font database”
Warning message in axis(side = side, at = at, labels = labels, ...):
“font family 'Georgia' not found in PostScript font database”
Warning message in axis(side = side, at = at, labels = labels, ...):
“font family 'Georgia' not found in PostScript font database”
[...]

PR #749, opened 2025-04-11
Let different cells use the same graphics device

### LINKED ISSUE #240: plot.new has not been called yet error message
I am working my way through DataCamp's ggplot2 tutorial. I keep getting the above error when I try to run this code.  It will run on their Console, and will run in RStudio, but will not run in Jupyter.  Any thoughts.  I get the "plot.new has not been called yet" error after the abline code, and the lapply code.

``` R
# Basic plot
mtcars$cyl <- as.factor(mtcars$cyl)
plot(mtcars$wt, mtcars$mpg, col = mtcars$cyl)

# Use lm() to calculate a linear model and save it as carModel
carModel <- lm(mpg ~ wt, data = mtcars)

# Call abline() with carModel as first argument and lty as second
abline(carModel, lty = 2)

# Plot each subset efficently with lapply
# You don't have to edit this code
lapply(mtcars$cyl, function(x) {
  abline(lm(mpg ~ wt, mtcars, subset = (cyl == x)), col = x)
  })
```

ma1112:
This PR resolves #240 , and as a side effect it also resolves issues reporting various behaviour when plotting with IRKernel, e.g. also #722.

The goal of this PR is to let different cells use the same graphics device. I.e. when one plots in one cell e.g. with the `plot` command and then enhances the plot in the following cell e.g. with `lines`, the cumulative effect of those two commands should be visible in the output of the former cell.

`IRKernel` calls `evaluate` to run R code cell by cell, and `evaluate` creates a new graphics device for each function call. See [the evaluate source code](https://github.com/r-lib/evaluate/blob/8d2bb1f353b878178b042d36ddcff2cca4a2be5e/R/evaluate.R#L77) 

With this change, `IRKernel` now creates its own graphics device and calls `evalute()` with `new_device = FALSE.`

Also, in order to capture plots properly, `IRKernel` now executes `dev.control(displaylist = "enable")`, similarly to `evaluate` that executes this for each kernel it creates ([code](https://github.com/r-lib/evaluate/blob/8d2bb1f353b878178b042d36ddcff2cca4a2be5e/R/watchout.R#L9C5-L9C40)). This is needed to capture the effects of multiple plotting functions applied on the same device.

By re-using the same graphics device, if one creates a plot in one cell, `evaluate` will call `handle_graphics()` in all subsesquent cells, even if the cell does not plot anything. To handle that, guardrails are applied in order not to send plots multiple times to the UI, in a from of introducing the boolean variable `last_recorded_plot_sent`.

Moreover, as a side effect, `IRKernel`'s own device settings are now taking effect, that has been hidden by `evalaute`'s default `new_device` option that has set the device to `pdf` for each execution. See the IRkernel code [here](https://github.com/IRkernel/IRkernel/blob/124f2347f15cfadeabc6738868e05e6087f4c456/R/environment_shadow.r#L121) and evaluate's code [here](https://github.com/r-lib/evaluate/blob/8d2bb1f353b878178b042d36ddcff2cca4a2be5e/R/watchout.R#L8).


R/execution.r
```diff
@@ -232,8 +250,14 @@ stream = function(output, streamname) {
...
 handle_graphics = function(plotobj) {
     log_debug('Graphics output...')
-    if (!plot_builds_upon(last_recorded_plot, plotobj)) {
+    if (plot_identical(last_recorded_plot, plotobj)){
+        log_debug('Plot is identical to the last one, not sending')
+        return()
+    }
+    if (!plot_builds_upon(last_recorded_plot,plotobj) && !last_recorded_plot_sent) {
+        log_debug('new plot does not build open the last plot. Sending unsent last plot.')
         send_plot(last_recorded_plot)
```
ma1112, line 259: To be honest I do not really understand what is the reason for sending the plot here, as well as after we call `evaluate()`, so I tried keeping the same logic with the new re-useable graphics devices.

R/execution.r
```diff
@@ -8,17 +8,34 @@ namedlist <- function() setNames(list(), character(0))
...
-plot_builds_upon <- function(prev, current) {
-    if (is.null(prev)) {
+
+plot_builds_upon <- function(prev, current, require_exact_length = FALSE) {
+    if(is.null(prev) && is.null(current)) {
         return(TRUE)
     }
+
+    if (is.null(prev) && !is.null(current)) {
+        return(!require_exact_length)
```
flying-sheep, line 18: can you explain the logic here?

ma1112: if `require_exact_length` is set to false, the function behaves as the former version of `plot_builds_upon`. Specifically, in this branch  `is.null(prev)` is true, so the function returns `TRUE`. 

if `require_exact_length` is set to true, the goal of the function is to check if the two plots are exactly the same. 
If `prev` is null but `current` is not null, they can not be the same and therefore we return `FALSE`. 

I am open to suggestions on how to rewrite the function to make it more readable. Maybe renaming `require_exact_length` to `require_exact_match` or explicitly  writing out the logic here would help more?
E.g. 

````
    if (is.null(prev) && !is.null(current)) {
        if(require_exact_length) { 
            return FALSE
        } else { 
            return TRUE
    }
```

ma1112:
@flying-sheep I know you said you had little time to review PRs but I'm wondering if you might find some time to look at this PR, or at least set some expectations on when you might have a chance to take a look.

R/execution.r
ma1112: @flying-sheep  does the explanation above help?

ma1112:
@flying-sheep  any chance you have some time to look at the PR this week?

